PR #381, opened 2018-12-27
Environment_oM: Issue373- Change order of Enums

### LINKED ISSUE #373: Environment_oM: Change order in Enums Result files
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

mariatjader:
To find the right types and units in a more user-friendly way the Enums are put in an A-Z order.
Fixes #373


al-fisher:
Before merging be good to have quick look at order of `ProfileResultUnits`.
Agree with your reorganising `enums` as alphabetical as a default, especially when the number of options is large. However for the `ProfileResultUnits` would sorting by size of time unit be more logical/intuitive though? Genuine question. Perhaps @michaldengusiak/@rwemay can you provide an additional opinion/comment?

Going forward - and perhaps one to include in future coding/UI style guidance - best practice of intuitiveness of `enums`:
Alphabetical should perhaps be a good default unless 
1. the type has a more intuitive order or 
1. an agreed frequency of use can not be defined/is not appropriate.

Otherwise  - in addition to addressing @michaldengusiak point above about breaking the `ProfileResultType` into three smaller more manageable result types - happy to merge

rwemay:
Enum ordering is a pain, but I think it would be good to establish some guidance. I guess alphanumeric should be a default sort order unless there is a more natural order (such as time/return period). I would be tempted to rename the ResultTypeUnit to something more like ResultPeriod or similar - to make it explicit (and any 'units' we want to deal with explicitly with the Units_Engine). 

One thing I had thought about in the guidance is adding some enum numbering placeholders. I've not done this so not sure how well it works (or if it works) but if we were to, by default, use 10, 20, 30 numbering (or more digits for enums we are worried will get much longer) it should allow us to add enums later without changing the originals. 

The other comment would be - have you looked at/spoken to @IsakNaslundBh  on the Result classes we've been using on the structures side? There are a few things to note about them: 
 - Use of immutable classes (IResult) to ensure that results that are pulled cannot be modified (only constructed and all relevant results/data specified at construction). 
 - Use of multiple result classes that inherit from IResult and result abstracts that allow for more intelligent 'sort' methods with IComparables on the abstract.

FraserGreenroyd:
Yes we spoke with Isak back in November over it and we agreed the results were too dissimilar at that time to be able to use the Results classes that he was using.

Personally I would recommend getting this merged in because neither @michaldengusiak nor @mariatjader are back until next week to be able to handle the additional queries. I have other priorities this week so won't be looking at it either, but not merging it in will start to hold things up. Thus, my recommendation would be to merge in now, and raise separate question/issue items to discuss around this. Either that or reject and close the PR so we don't have anything holding things up for a week?

rwemay:
@FraserGreenroyd , I'll merge this now if it's holding you up - relatively simple enum changes, hopefully won't break too many scripts with the reordering - assume you guys know the impacts. We should have a review later on the result format - use of classes/enums, and enum numbering (to allow later modification by using stepped numbering). Could you raise another issue so we can plan in?

PR #387, opened 2019-01-11
Add elementID property to BuildingElement

### LINKED ISSUE #386: Environment_oM: Add ElementID property to BuildingElement
Following discussion with @michaldengusiak and @ZiolkowskiJakub to allow some pushing toolkits to be agnostic, we should have a `string` property on the `BuildingElement` which references the `CADElementID`.

FraserGreenroyd:
Fixes #386

### DIFF
--- a/Environment_oM/Elements/BuildingElement.cs
+++ b/Environment_oM/Elements/BuildingElement.cs
@@ -39,6 +39,8 @@ namespace BH.oM.Environment.Elements
         public ICurve PanelCurve { get; set; } = new PolyCurve();
         public List<Opening> Openings { get; set; } = new List<Opening>();
 
+        public string ElementID { get; set; } = ""; //ElementID to reference original CAD geometry
+
         /***************************************************/
     }
 }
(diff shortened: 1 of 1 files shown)


rwemay:
Should this not be in CustomData as it's information imprinted from an adapter? (Also, I don't think we should be saying 'CAD' anywhere in the BHoM since it should be agnostic.

FraserGreenroyd:
@rwemay the property is just `ElementID` not `CADElementID` to avoid that link to specific CAD products - just the comment gives explanation for its use temporarily for people on the open source coming to it and wondering what it might be for for any Computer Aided Design package if that makes sense?

We are moving away from relying so heavily on custom data particularly in the XML Toolkit due to the ongoing discussions as to whether or not to keep custom data. This helps us become more independent from Revit as well, as this data was previously stored as `Revit_elementId` in the custom data so follows the discussion with @michaldengusiak and @ZiolkowskiJakub from Friday to make this agnostic.

FraserGreenroyd:
@rwemay I have removed the comment entirely to avoid confusion then, and will drop you an email regarding the rest of it to keep the GitHub area clean :smile:

Issue #393, opened 2019-01-16
Common_oM: Material


ZiolkowskiJakub:
Hi,

I would like start here s discussion about Materials and how we can make common approach for Material definition.

Currently there are separate definitions for Environmental and Structural material.

We should discuss where Material definition goes. Is it Discipline specific? If yes then why structural material is in Common namespace?

Is it possible to implement core Interface or all Materials? As far as I can see there are some common properties.

rwemay:
Hi @ZiolkowskiJakub , material definitions should be in the common namespace, and definitely discipline agnostic.

michaldengusiak:
I would think we have some:
-  first level default basic properties common among all discipline 
- second level discipline specific we need to think how we divide and store data need to daylight (more basic) and energy simulation 
- and last third level  software specific variant for above 

I guess would be good to Sketch some example

1. Name= Concrete Color:  hex value definition
2. **Daylight**: Refelctance Internal,External;  Transmission Internal,External   **EnergySimulation**: Conductivity, SpecificHeat, Density, VapourDiffusionFactor, SolarReflectanceExternal, SolarReflectanceInternal, LightReflectanceExternal, LightReflectanceInternal, EmissivityInternal, EmissivityExternal, Roughness    

3. Software Specific **TAS**: IgnoreInUvalue, IsBlind or **Radiance**:

![image](https://user-images.githubusercontent.com/26113670/51303857-842efb00-1a2e-11e9-87ab-e1ea715353fa.png)

so 

RGB inoput could be recalculated form Color:  hex value definition [see example here](https://stackoverflow.com/questions/12155811/how-to-convert-hexadecimal-ffffff-to-system-drawing-color)

```
## ----------------------- Concrete ----------------------- ##
#
# concrete / (Color_45058_46543_46996)
# 
dusty_wide plastic (Color_45058_46543_46996)
0
0
5 0.687541 0.710201 0.717113 0 0
```

IsakNaslundBh:
@rwemay Might be worth considering having all the base and common parameters on the base material class, such as density, and then have like different modules on this class covering the different needs, such as structural, env, acoustic etc?

Given that we will need to expand our structural material capabilities to for example orthotropic materials, and also probably make a divide between the isotropic materials for design properties, it feels like it can get a bit out of hand to have it all on the base class.

Assume the case is the same for env.

Basically what @michaldengusiak described above.

rwemay:
[...]

PR #398, opened 2019-01-24
WIP on #293 - extended properties for discussion

### LINKED ISSUE #293: BHoM: Add support for ECS to the base BHoMObject
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

FraserGreenroyd:
Fixes #293 

Begins work to include extended properties into base BHoM object for all BHoM objects to then make use of as we look to use extended properties on some objects. This PR has been raised to begin the review/discussion of how to implement this and showcase the ideas following discussions with @al-fisher  and @rwemay .

### Test files
N/A

### Changelog
####Added
 - Added `ExtendedProperties` functionality

### DIFF
--- a/BHoM/BHoM.csproj
+++ b/BHoM/BHoM.csproj
@@ -57,10 +57,12 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="BHoMExtendedProperties.cs" />
     <Compile Include="CRUD\IExecuteParams.cs" />
     <Compile Include="CustomObject.cs" />
     <Compile Include="CRUD\CrudConfig.cs" />
     <Compile Include="CRUD\Enums\PushActionType.cs" />
+    <Compile Include="Interface\IBHoMExtendedProperties.cs" />
     <Compile Include="Interface\IBHoMGroup.cs" />
     <Compile Include="Interface\IObject.cs" />
     <Compile Include="Interface\IImmutable.cs" />
--- a/BHoM/BHoMExtendedProperties.cs
+++ b/BHoM/BHoMExtendedProperties.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * This file is part of the Buildings and Habitats object Model (BHoM)
  * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
  *
@@ -25,18 +25,12 @@ using System.Collections.Generic;
 
 namespace BH.oM.Base
 {
-    public interface IBHoMObject : IObject
+    public class BHoMExtendedProperties : IBHoMExtendedProperties
     {
-        Guid BHoM_Guid { get; set; }
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
 
-        string Name { get; set; }
-
-        List<IBHoMExtendedProperties> ExtendedProperties { get; set; }
-
-        HashSet<string> Tags { get; set; }
-
-        Dictionary<string, object> CustomData { get; set; }
-
-        IBHoMObject GetShallowClone(bool newGuid = false);
+        public string ElementID { get; set; } = "";
     }
 }
--- a/BHoM/BHoMObject.cs
+++ b/BHoM/BHoMObject.cs
@@ -35,6 +35,8 @@ namespace BH.oM.Base
 
         public string Name { get; set; } = "";
 
+        public List<IBHoMExtendedProperties> ExtendedProperties { get; set; } = new List<IBHoMExtendedProperties>();
+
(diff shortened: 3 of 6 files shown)


adecler:
Glad to see some action happening around ECS 👍 .
A few initial comments:
- I am not convinced we need an `ElementID` property (especially a string) and therefore a `BHoMExtendedProperties` class (an interface would suffice).
- Names like `BHoMExtendedProperties` and `BuildingElementExtendedProperties` are really long and technical. I wonder if we could use something a bit less so. If you look at the `Human_oM` ([here](https://github.com/BHoM/BHoM/blob/master/Humans_oM/Interfaces/IHumanRole.cs), [here](https://github.com/BHoM/BHoM/blob/master/Humans_oM/Employee.cs), and [here](https://github.com/BHoM/BHoM/blob/master/Humans_oM/Skeleton.cs)), you will see I had already started a similar light ECS infrastructure with the concept of `Role` (only `IHumanRole` used at the moment). The idea of the name is that you attach extra functionality to a possibly varied set of objects so you give extra capability/purpose/role to that object. Not necessarily the most adequate name for a general purpose but something in between `BHoMExtendedProperties` and `Role` would be nice.
- There is a need for a discussion around the parts of ECS we want to embrace and the ones we want to leave behind. More specifically, the object definition aspect is only the tip of the iceberg while there is a good number of technical aspects to navigate around the handling (accessing, setting, updating,...) of those extended properties/roles.

IsakNaslundBh:
Not exactly the same, but linked to ECS, so might be something to keep in mind as well are the discussions started here:

https://github.com/BHoM/BHoM/issues/393

FraserGreenroyd:
@IsakNaslundBh yes good to link that in though I'd hope we could move this one forward quicker as it's holding up things from BEnv side whereas I'm not sure if materials is holding anything up for either side? Correct me if I'm wrong though cause I can't guess for your side :smile:

@adecler I have removed the `BHoMExtendedProperties` file and removed the `ElementID` from the interface (and moved it to the `BuildingElementExtendedProperties` for now). I haven't renamed anything yet as am waiting to see if anyone has any good ideas on how to name it :smile:

epignatelli:
What about `EnvironmentAware` ?

IsakNaslundBh:
@FraserGreenroyd no rush at all from the structures side atm, so no reason to hold this up for that sake! Just linking over as the principle overlap!

al-fisher:
@FraserGreenroyd @michaldengusiak As discussed suggesting you remove the `IBHoMExtendedProperties` and the Implementation on the `BH.oM.Base.BHoMObject` - such that this PR can therefore be a prototype of ECS simply for the `BuildingEnvironmentObjects`.

This will allow you to proceed and prototype in parallel to the `Human` example.   
And I would suggest @IsakNaslundBh we also discuss implementing a similar (or not) ECS approach for `Material`.

There are a number of effects and subtleties that will be uncovered - we will learn a lot.
In addition some further considerations will need to be considered around how to expose and reflect all ECS properties and behaviours in UI etc. Good to test in discipline code under specific use-cases before working to generalise and mapping to parent BHoM object and all dependant objects.

@adecler does this make sense to you too?

al-fisher:
Still struggling to think of the right term - somewhere between Role and ExtendedProperties although I do like what you are trying to achieve with BuildingAware @epignatelli 
Will keep thinking

FraserGreenroyd:
@al-fisher changes made as discussed :smile:

al-fisher:
Summary of discussions with @FraserGreenroyd @michaldengusiak @ZiolkowskiJakub -

@FraserGreenroyd _et al._ to post a first draft of the list (maybe just part of?! as previously super long) of analysis/design properties broken up and categorised into "Extended Properties" modules. Grouping the properties based on themes. This will help to flesh out the immediate BEnv implementation of an ECS prototype.

Also from discussion likely the name `BuildingElementExtendedProperties` will be changed / such a generic object removed - with then a number of `ExtendedProperties` objects created more specifically named around the classification of Properties it is containing. i.e. `CoolingLoadExtendedProperties` or `TemperatureExtendedProperties` etc.

Cheers all - great chat 😄 Please check the above makes sense/accurate - can follow up tomorrow too perhaps. 
@rwemay FYI

FraserGreenroyd:
As an FYI for all involved - this PR won't rebase nicely following changes made in PR #405 due to conflicts in the `Environment_oM.csproj` file. I'll be resolving this once @michaldengusiak finishes the diagram for us all but just to give everyone a heads up!

PR #409, opened 2019-02-04
Environment_oM: Extended properties for Environmental objects

### LINKED ISSUE #408: Environment_oM: Critical extended properties prototype
This issue is for the critical extended properties prototyping being done in Environment_oM. @michaldengusiak to confirm the list of critical properties. Further issues will be made for the less-critical properties for further discussion as we progress.

Related to #293 & #398

FraserGreenroyd:
Fixes #408 

Adds the critical extended properties prototyping for Environmental Objects

### Test files
N/A

#### Added
 - Added beginning Extended Properties (meta, analytical, and results) to Environmental Objects

### Additional comments
See #398 for more information regarding this work.


Environment_oM/Elements/Building.cs
```diff
@@ -29,7 +29,7 @@
 
 namespace BH.oM.Environment.Elements
 {
-    public class Building : BHoMObject, IBuilding
+    public class Building : BHoMObject, IBuildingObject
     {
```
al-fisher: Will you dispense with or clarify the distinction of `IBuilding` and `IBuildingObject` going forward? Is a separate issue to this `ExtendedProperties` prototype but now we have an `IBuilding` interface not implemented by our `Building` class.

Environment_oM/Properties/BuildingElementMetaProperties.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Environment.Interface;
+
+namespace BH.oM.Environment
+{
+    public class BuildingElementMetaProperties : IBHoMExtendedProperties
+    {
```
al-fisher: Naming is really hard! - but wonder if we can think of a more suitable name than `Meta` - as a user will struggle to know when I will need/use these or not.

al-fisher: How are these properties being grouped?

Environment_oM/Properties/IBHoMExtendedProperties.cs
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+using System;
+using System.Collections.Generic;
+
+using BH.oM.Base;
+
+namespace BH.oM.Environment.Interface
+{
+    public interface IBHoMExtendedProperties : IObject
```
al-fisher: This will at some point likely need to be renamed! But as none of us can think of a better name right now 😄 - happy to plough on!
It will also potentially/hopefully merge will a more generally applied approach to "Extended Properties" so plenty of opportunities to refactor

Environment_oM/Properties/BuildingElementMetaProperties.cs
FraserGreenroyd: Currently they're being grouped by `meta` (properties that describe the object but do not define it), `analytical` (properties that the object needs for its analytical representation but not physical representation) and `results` (properties that the object has for future analysis from some result of another 'thing' - though this is the one that is not quite defined fully yet and will be looked at in ongoing prototyping).
So for example, `Temperature` would be an `analytical` extended property of any object which then has the capacity to store them. I hope that makes sense? :smile:

Environment_oM/Elements/Building.cs
FraserGreenroyd: `IBuilding` is likely to become deprecated/removed as I found it wasn't doing anything at this stage whereas `IBuildingObject` is being used to group our environmental objects within Engine work, etc. if that makes sense

FraserGreenroyd:
@al-fisher have changed `meta` to `context` but still not sure whether that's right - what do you think (I've put a comment around your comment with the definition of each group currently for context) :smile:

al-fisher:
Thanks for this - specifically on `context` rather than `meta` - I do think that is an improvement. Will approve so you can proceed to prototype and test further.
Cheers

PR #414, opened 2019-02-11
Modified InternalGains and added Interfaces to different internal gai…

### LINKED ISSUE #413: Environment_oM: Agnosticise internal gains objects
In order to move into multi-software dynamic thermal modelling we need to make our internal gains definitions more software agnostic. 

I thought I'd dive right in and try doing this myself on a new branch - which will appear very shortly.

tg359:
…ns types - (this breaks the TAS toolkit)

Fixes #413 

`InternalGain` objects for `People`, `Lighting` and `Equipment` added and modified from original `InternalGain` object. 

Each has a `Profile` to define when the gain is present/at what level, and a value to be assigned to that profile with a metric describing what that value represents (e.g. `NumberOfPeople`/`Watts`/`WattPerSquareMetre`).

Additional `enums` and `Interfaces` also included to enable inheritance from default gains types, as well as additional properties such as illuminance and lighting efficacy for lighting gains.

_**THIS BREAKS `TAS_Toolkit` due to the fact that it relies on the `InternalConditions` object, which uses an `InternalGain` object. This will need to be addressed during review!!!!**_

#### Changed:
- `InternalGain` InternalGain object modified in the Environment_oM for the InternalGains class

#### Added:
- `InternalGainMetric` InternalGainMetric enum added in the Environment_oM for the InternalGains class
- `InternalGainType` InternalGainType enum added in the Environment_oM for the InternalGains class
- `IInternalGain` IInternalGain interface added in the Environment_oM for the InternalGains class
- `IInternalGainProperties` IInternalGainPropertiesinterface added in the Environment_oM for the InternalGains class
- `IInternalGainPropertiesEquipment` IInternalGainPropertiesEquipment added in the Environment_oM for the InternalGains class
- `IInternalGainPropertiesLighting` IInternalGainPropertiesLighting added in the Environment_oM for the InternalGains class


FraserGreenroyd:
@tg359 can you add the change log details for this PR please? BHoM_Bot will scrape the PR note rather than commit history. I'd think this is something like:

#### Added
 - Added InternalGain properties

etc. etc.

tg359:
> @tg359 can you add the change log details for this PR please? BHoM_Bot will scrape the PR note rather than commit history. I'd think this is something like:
> 
> #### Added
> * Added InternalGain properties
> 
> etc. etc.

Changed the PR description to include these. Some ambiguity around `####Added`, where I'm not adding a file in some cases, but instead modifying one. Is there anywhere describing other tags available (e.g. `####Modified`)? I checked the link given but nothing there yet.

Environment_oM/Elements/InternalGain.cs
```diff
@@ -28,24 +28,15 @@
 
 using BH.oM.Base;
 using BH.oM.Environment.Properties;
+using BH.oM.Environment.Elements;
+using BH.oM.Environment.Interface;
 
-namespace BH.oM.Environment.Elements
+namespace BH.oM.Environment.InternalGains
```
FraserGreenroyd: I think we should generalise this namespace as we should, eventually, be getting external gains as well and I personally don't think we need separate namespaces for both.

I recommend changing this to `Gains` as the namespace (`BH.oM.Environment.Gains`) rather than `InternalGains`

Environment_oM/Elements/InternalGain.cs
```diff
@@ -28,24 +28,15 @@
...
 {
-    public class InternalGain : BHoMObject
+    public class InternalGain : BHoMObject, IInternalGain
```
FraserGreenroyd: I don't believe we need this `IInternalGain` interface. However, we might want a `Gain` interface for when we eventually get External Gain objects as well.

Rename to `IGain`.

Environment_oM/Elements/InternalGain.cs
```diff
@@ -28,24 +28,15 @@
...
     {
-        /***************************************************/
-        /**** Properties                                ****/
-        /***************************************************/
-        public List<Profile> Profiles { get; set; } = new List<Profile>();
+        public InternalGainType InternalGainType { get; set; } = InternalGainType.Undefined;
```
FraserGreenroyd: Similar, I think `Internal` isn't right here when we later want external gains, and I can't think that an External Gain is going to need different properties to the internal one. So let's remove all references to `InternalGain` and replace with `Gain` going forward.

Environment_oM/Enums/InternalGainMetric.cs
```diff
@@ -0,0 +1,34 @@
...
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.Environment.Elements
```
FraserGreenroyd: Put this Enum in the same namespace as the element. Also put it in the same folder structure please.

Environment_oM/Enums/InternalGainType.cs
```diff
@@ -0,0 +1,32 @@
...
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.Environment.Elements
```
FraserGreenroyd: Put this Enum in the same namespace as the element. Also put it in the same folder structure please.

Environment_oM/Enums/InternalGainType.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public enum InternalGainType
```
FraserGreenroyd: Change to just `GainType`
```suggestion
    public enum GainType
```

Environment_oM/Enums/InternalGainMetric.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public enum InternalGainMetric
```
FraserGreenroyd: Change to just `GainMetric`
```suggestion
    public enum GainMetric
```

Environment_oM/Interface/IInternalGain.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Environment.Elements;
+
+namespace BH.oM.Environment.Interface
+{
+    public interface IInternalGain: IBHoMObject
```
FraserGreenroyd: Change to just a `IGain` and a space before the colon.
```suggestion
    public interface IGain : IBHoMObject
```

Environment_oM/Interface/IInternalGain.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        IInternalGainProperties InternalGainProperties { get; set; }
+    }
```
FraserGreenroyd: I would say all Gain objects should also have their type contracted in by the interface.

Environment_oM/Interface/IInternalGainProperties.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Environment.Elements;
+
+namespace BH.oM.Environment.Interface
+{
+    public interface IInternalGainProperties : IBHoMObject
```
FraserGreenroyd:
```suggestion
    public interface IGainProperties : IBHoMObject
```

Environment_oM/Properties/InternalGainPropertiesEquipment.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Environment.Interface;
+using BH.oM.Environment.Elements;
+
+namespace BH.oM.Environment.Properties
+{
+    public class InternalGainPropertiesEquipment : BHoMObject, IInternalGainProperties
```
FraserGreenroyd:
```suggestion
    public class GainPropertiesEquipment : BHoMObject, IlGainProperties
```

Environment_oM/Properties/InternalGainPropertiesLighting.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Environment.Interface;
+using BH.oM.Environment.Elements;
+
+namespace BH.oM.Environment.Properties
+{
+    public class InternalGainPropertiesLighting : BHoMObject, IInternalGainProperties
```
FraserGreenroyd:
```suggestion
    public class GainPropertiesLighting : BHoMObject, IGainProperties
```

Environment_oM/Properties/InternalGainPropertiesPeople.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Environment.Interface;
+using BH.oM.Environment.Elements;
+
+namespace BH.oM.Environment.Properties
+{
+    public class InternalGainPropertiesPeople : BHoMObject, IInternalGainProperties
```
FraserGreenroyd:
```suggestion
    public class GainPropertiesPeople : BHoMObject, IGainProperties
```

Environment_oM/Elements/InternalCondition.cs
```diff
@@ -43,7 +44,7 @@ public class InternalCondition : BHoMObject
 
         public List<SimulationDayType> DayTypes { get; set; } = new List<SimulationDayType>();
 
-        public InternalGain InternalGain { get; set; } = new InternalGain();
+        public Gain InternalGain { get; set; } = new Gain();
```
FraserGreenroyd:
```suggestion
        public Gain Gain { get; set; } = new Gain();
```

tg359: Damn - I thought I got all of those! Fixed in the next commit.

tg359:
@al-fisher The `GainUnit` here is a bit of an odd one as it relates very specifically to the _non-standard_ units used for `InternalGains`. We've had success using the `UnitsNet` package in the `Climate_Toolkit` so I reckon it would be easiest to roll-out into other toolkits to save on unecessary rework. 

There may be fringe cases such as the unit `CubicMetrePerSecondSquareMetreExternal` here (the amount of air leakage through external walls - which isn't a standard unit in the `UnitsNet` library) that may need some thought - perhaps extending `UnitsNet` within the `Units_Engine` for those cases?

PR #427, opened 2019-03-05
Geometry_oM: Creation of PlanarSurface

### LINKED ISSUE #424: Goometry_oM : Creation of PlanarSurface
As discussed and agreed https://github.com/BHoM/BHoM/issues/422#issuecomment-469346777 - useful to create a simple planar surface definition - defined by bounding planar curves.

IsakNaslundBh:

Closes #424

Adding PlanarSurface object

#### Added:
- PlanarSurface class added, defined by external and internal EdgeCurves

### DIFF
--- a/Geometry_oM/Surface/PlanarSurface.cs
+++ b/Geometry_oM/Surface/PlanarSurface.cs
@@ -32,10 +32,10 @@ namespace BH.oM.Geometry
         /***************************************************/
 
         [Description("Closed planar curve defining the outer boundary of the surface")]   
-        public ICurve ExternalEdge { get; set; } = null;
+        public ICurve ExternalBoundary { get; set; } = null;
 
         [Description("List of closed (co)planar curves defining any internal openings in the surface")]
-        public List<ICurve> InternalEdges { get; set; } = null;
+        public List<ICurve> InternalBoundaries { get; set; } = null;
 
         /***************************************************/
     }
--- a/Geometry_oM/Surface/PlanarSurface.cs
+++ b/Geometry_oM/Surface/PlanarSurface.cs
@@ -35,7 +35,7 @@ namespace BH.oM.Geometry
         public ICurve ExternalBoundary { get; set; } = null;
 
         [Description("List of closed (co)planar curves defining any internal openings in the surface")]
-        public List<ICurve> InternalBoundaries { get; set; } = null;
+        public List<ICurve> InternalBoundaries { get; set; } = new List<ICurve>();
 
         /***************************************************/
     }
--- a/Geometry_oM/Surface/PlanarSurface.cs
+++ b/Geometry_oM/Surface/PlanarSurface.cs
@@ -1,0 +1,42 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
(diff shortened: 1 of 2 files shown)


Geometry_oM/Surface/PlanarSurface.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace BH.oM.Geometry
+{
+    public class PlanarSurface : ISurface
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Closed planar curve defining the outer boundary of the surface")]   
+        public ICurve ExternalEdge { get; set; } = null;
```
FraserGreenroyd: Should this not also be `Edges` plural? Surely a planar surface has more than one outer edge defining it? I know it'll be a singular curve and not a collection, but the variable name suggests it might only be one edge of the external boundary. Suggestions from me to start discussion cause I'm not sure what it could be (only know that singular edge doesn't sound right):

 - ExternalEdges
 - BoundingCurve
 - ExternalBoundary

IsakNaslundBh: Could update the name like that. Reason for naming it in singular is that it should be A `Circle` or A `Polycurve` etc. 

Would not like it to be called `ExternalEdges`, think that is more confusing tbh. Would be happy with `ExternalBoundary` and `InternalBoundaries` though. Methods implemented in the engine so far call the edge curves Edges though, so probably good to keep it consistent.

FraserGreenroyd: Yeah I've just looked at the engine and seen that. I think `ExternalBoundary` and `InternalBoundaries` would make sense and be consistent together.

IsakNaslundBh: @al-fisher @rwemay @pawelbaran Any comments on this?

epignatelli: Sorry, I didn't get the reason we're not happy with the current one, what is it? 

If it's necessary, happy for it to be changed, but I need to change the Grasshopper and the Rhinoceros toolkit if this happens - I'd like to avoid that if not necessary.

pawelbaran: I would strongly disagree with `ExternalEdge`. It is confusing as one can think of a single edge, which of course is not true / illogical. I also get @IsakNaslundBh point in avoiding  `ExternalEdges`, which, in my opinion, brings us to the most sensible option which is `ExternalBoundary` and `InternalBoundaries`.

I would even dare to consider our `PlanarSurface` as a planar brep defined by a set of outlines - we have no ability to define or analyse it based on UV, so in my understanding it is not a true surface by definition. In such case Boundary seems to be the only appropriate word for the outlines.

al-fisher: I would agree that boundary is the most appropriate term of the two actually (sorry @epignatelli if that means changes 😩)
So `ExternalBoundary` and `InternalBoundaries`

IsakNaslundBh: ok, thanks all. WIll update, and can cover rhino/GH as well as it just considers a simple property name update.

@ZiolkowskiJakub FYI in case you have started working on this on the revit side.

epignatelli: ahahah I can definitely do that :D 

@pawelbaran So what did we settle for? `PlanarBoundary` , just `Boundary`? And for the properties?

IsakNaslundBh: Have update oM, Engine and Rhino now. Does not seem to require any change to GH.

ZiolkowskiJakub: Hi @IsakNaslundBh,

No worries about me. I will start updating Revit_Toolkit when this PR is merged.

PR #423, opened 2019-03-04
Environment_oM - Remove deprecated properties from BuildingElement

FraserGreenroyd:
**DO NOT MERGE UNTIL ALL DEPENDENT PRS ARE HAPPY** 

### Issues addressed by this PR
Fixes #420 

Removes `BuildingElementProperties` class
Removes `ElementID` from `BuildingElement`

Both now served through ExtendedProperties current prototype

### Test files
N/A

#### Removed
 - Removed `BuildingElementProperties` class
 - Removed `BuildingElementProperties` property from `BuildingElement`
 - Removed `ElementID` property from `BuildingElement`

### Additional comments
A hard task to remove elements and update all the relevant TKs beneath the core oM...

PR #428, opened 2019-03-05
Environment_oM - material updates

### LINKED ISSUE #390: Environment_oM: Add MaterialProperties
Material MaterialPropertiesTransparent
missing `IsBlind`

Material MaterialPropertiesOpaque
missing `IgnoreInUvalueCalcs`

### LINKED ISSUE #417: Environment_oM: Material roughness
Materials should also specify their roughness in addition to a construction roughness. This is to be looked at in either March or April sprints pending the outcome of other material refactoring.

FraserGreenroyd:
Fixes #390 
Fixes #417 

`IsBlind` has not been added despite being requested in #390 - instead a material function has been added to define what function the material is playing in the overall construction. This is more agnostic and better for the Environment_oM overall.

### Test files
N/A

#### Added
 - Added `MaterialFunction` property to `Material` for Environments
 - Added `MaterialRoughness` property to `Material` for Environments

### Additional comments
While ongoing prototyping is done with materials by @IsakNaslundBh , this work allows a move forward with materials and constructions in TKs being developed within BEnv by @tg359 and the gang.

### DIFF
--- a/Environment_oM/Enums/Materials/MaterialFunction.cs
+++ b/Environment_oM/Enums/Materials/MaterialFunction.cs
@@ -1,0 +1,44 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.Environment.Materials
+{
+    public enum MaterialFunction
+    {
+        Undefined,
+        Aggregate,
+        AirGap,
+        Asphalt,
+        Blind,
+        Board,
+        Brick,
+        Carpet,
+        Concrete,
+        Glass,
+        Insulation,
+        Metal,
+        Plaster,
+        Screed,
+        Tile,
+        Timber,
+    }
+}
--- a/Environment_oM/Enums/Materials/MaterialRoughness.cs
+++ b/Environment_oM/Enums/Materials/MaterialRoughness.cs
@@ -1,12 +1,35 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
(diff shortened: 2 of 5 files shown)


michaldengusiak:
this material function is very e+ specific and wwe should implement  a software specific place in `material extended property`

FraserGreenroyd:
Incorrect, this Material Function is deliberately agnostic. Materials do not need to know anything about the software they come from/go, so extended properties with a software place is not valid. Their data is agnostic and that's what this is, an agnostic way of determining the function of the material. This is in place of `IsBlind` which is solely TAS specific and provides a wider range of functionality than a single boolean.

michaldengusiak:
I disagree.. 
why I need to know if sth is `Brick `or `Asphalt`?

FraserGreenroyd:
Why do you need to know if it's `IsBlind` then by the same logic? This is the solution to the issue you raised in #390 ?

al-fisher:
Hi @FraserGreenroyd @michaldengusiak like the intent here.
Agree we need to avoid a situation where we need to implement many `Is"X"` boolean queries.

However good to get (for me anyway) some greater clarity over the definition and distinction between the object Type, its Usage and its Material/Construction.

We currently have following enums:

``` C#
MaterialType
ConstructionType
BuildingElementType
MaterialFunction 
MaterialRoughness
```

Which currently there is both potentially some overlap in their definitions and also some interdependence.

Can you share an example, for perhaps a couple of Building element Types say (Wall, Floor, Window, Blind) how each of the above are used. i.e. set or queried.

Will perhaps help to consolidate.

al-fisher:
A simple example GH script creating and then Exploding and Querying the above would perhaps be a good way of explaining, alongside a diagram... hope makes sense

michaldengusiak:
ok great we have discussion ...let me post what we just finish right now related to this subject
after hours of research and _discussion_ we achieved compromise with @tg359 for `MaterialFunction `
here is latest list:

MaterialFunction | Description
-- | --
Undefined, | Unknown or Error
Blind, | blinds,Lamellen, Jalousie only
Board, | Building panels boards, Bauplatten, Wandbauplatten
Conrete | Conrete, Grossformatige Bauteile
Masonry | Brick, Stones, Mauerwerk
Insulation, | Thermal insulation, Wärmedämmstoffe
Covering | Plaster, Screed, Mortal, Putze, Estrich, Mortel+Membrane, Carpet,   Asphalt, Tile, Beläge, Abdichtung
Timber | Timber, Timber Panels,   Plywood, Holzwerkstoffe
Other | Gas, Glass or Anything does   not fit above ie..standards Material

FraserGreenroyd:
@al-fisher to confirm, items like `ConstructionRoughness` and `ConstructionType` will be disappearing - but in a similar manner to #423 I need to provide the replacement and ensure all the dependent toolkits can update reliably before removing from the core oM and cause a massive backlog/breaking changes. Despite best efforts last September, the current oM has been proven to not be agnostic enough for the large range of TKs now being developed/worked on. This is part of the March sprint for BEnv to rectify this in order to push forward on our TKs, and will be a continued on going focus over the next few months.

@tg359 will a link to the existing workflow in Honeybee that we are trying to replicate. I am currently against doing any further work to a dependent TK on this due to the length of time lost in waiting for these types of PRs to be merged, and because there is no requirement for the Environment_Engine to make use of this functionality at this stage.

tg359:
Hi @al-fisher. Of the attributes above there are some pretty big differences. I've tried to justify these below:
- MaterialType - The highest level category for a material. It would either be a Gas (air, argon, krypton, CO<sub>2</sub>), Transparent (a material which lets solar, visible and infrared radiation through), Opaque (a fully opaque material). Common to pretty much all thermal modelling tools, we cannot have an opaque material in a window construction as this causes a simulation to fail.
- MaterialFunction - The next level down. This further enables the user to choose materials to make-up a construction. We would of course have pre-defined constructions comprised of materials (meeting certain design standards), but the ability to define custom constructions would require some additional sorting to find an 'insulation' material to add to a wall, or a 'covering' material to use for a roof.
- MaterialRoughness - Imagine hot air rising up a wall, and the turbulence of that air impacted by the roughness of that material. As a property of a material this changes the heat transfer into that air interpreted by the heat transfer method being used.
- ConstructionType - This is going, as it’s essentially the same as material type - but material type also includes gas)
- BuildingElementType - This is the type of the building element, i.e. wall, floor, ceiling, etc. A physical component of the building that can construct it. We can't construct a building with an opening (a door/window) and need a building element to host that opening. @FraserGreenroyd please correct this if I've misinterpreted.
---------
As for a simple script, we're essentially at the stage where we're porting a lot of what Honeybee does into BHoM for platform agosticity. Here are some links to scripts which in a very basic way create and modify materials and constructions for an EnergyPlus simulation which do all of the above (in a similar, but not identical way).
- http://hydrashare.github.io/hydra/viewer?owner=chriswmackey&fork=hydra_2&id=Full_Building_Energy_Simulation&slide=0&scale=1&offset=0,0
- http://hydrashare.github.io/hydra/viewer?owner=chriswmackey&fork=hydra_2&id=Green_Roof_In_Energy_Model&slide=0&scale=1&offset=0,0
- [GH file for window constructions](https://github.com/BHoM/BHoM/files/2935362/FurtherTesting3_CWM2.zip)

FraserGreenroyd:
I am now closing and deleting this PR/branch because the changes made are being incorporated in the #457 branch as part of the host of oM upgrades.

PR #430, opened 2019-03-07
Update COPYRIGHT_HEADER.txt date

### LINKED ISSUE #429: BHoM : Update date in COPYRIGHT_HEADER.txt

al-fisher:

Closes #429 

Updating the stated date in the template copyright statement file to 2019

### DIFF
--- a/COPYRIGHT_HEADER.txt
+++ b/COPYRIGHT_HEADER.txt
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.


adecler:
copyrights are not exactly my area of expertise but it seems that having to update the copyright end date across all files every year is a bit of a waste. Granted, the bot can take care of it but still. 

So what is stopping us from setting the end date to something far in the future? A quick googling gives me different answers as to the legal duration of a copyright but I have things like 70, 95 and 120 years).

al-fisher:
Yes - the best strategy for this specific issue was discussed last year with our copyright and software licencing lawyer. A final approach was not agreed, as there are the couple of issues you raise - minimising maintenance vs validity of statement.

Regardless of the date we agree - it will be part of our continuous integration and compliance to validate that the correct date (and entire copyright statement) is stated.

As you say we can then decide whether we a) do a bulk update of all code each year, b) update each file at the point of a modification and the PR (as the copyright statement is stating the period during which that particular file was created and modified), or finally, your proposal, c) bulk apply a future date.

The reason for updating this file now is - it gives us 9 months to decide the final strategy on the above. At least for brand new code that is being added. 
Let's aim to do sooner though 😉

PR #451, opened 2019-04-09
Physical/Analytical oM + Structure/Environment oM - MOCK UP

FraserGreenroyd:
This PR is raised as a discussion PR to allow any individual to comment or feedback on the prototyping work done by @IsakNaslundBh and myself over the last week or so.

This PR is not to be rebased at any time, as this is for discussion only. Subsequent PRs will be raised to merge in the elements of this PR along the implementation plan. More information about the order of things being done can be found [here](https://github.com/BHoM/BHoM/projects/1).

This prototype brings in the following projects:
 - Physical_oM - for any real-world object, expanding on the prototype done by @ZiolkowskiJakub previously
 - Analytical_oM - for the analytical representation of a real-world object.

Analytical_oM is designed to capture analytical properties that are joint between various disciplines (currently Structures and Environments), so common properties can be found here. It is using generics for the implementations down the line - these will be handled by a later Analytical_Engine for querying, etc.

The Structure_oM and Environment_oM demonstrate how @IsakNaslundBh and I intend to implement the new oM across our disciplines, and people developing for those disciplines are invited to review and provide feedback as you see fit.

If anyone has any questions, please post them here for visibility and @IsakNaslundBh or I will try our best to answer them.

Cheers!

Because I can only select 15 reviewers, these tags are to notify people I can't include on the right hand side:
@adecler 
@al-fisher 
@rwemay 
@tg359 
@ZiolkowskiJakub


Physical_oM/Properties/IFramingElementProperty.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+
+namespace BH.oM.Physical.Properties
+{
+    public interface IFramingElementProperty : IBHoMObject
```
MartinHenriksen: would it not make sense to add common properties to the interface? This would allow one to get e.g. ISectionProperty and orientation without having to cast the IFramingElementProperty to what ever it is implemented as. 
Maybe I lack imagination but I cannot think of a case where a FramingElement will not have a SectionProperty

FraserGreenroyd:
Just to update everyone tagged, the Environment oM changes are now PRed properly [here](https://github.com/BHoM/BHoM/pull/458).

Physical_oM/Properties/IFramingElementProperty.cs
IsakNaslundBh: Sorry for this late reply @MartinHenriksen , but:

Reason why there are no properties on the base interface is that you will have cases where you will have:
1. Varying crossection along the beam, for example a taper -> needing a collection of sections
1. For some really weird cases might have to resort to a Brep of the whole thing, if there are not possible to describe the framingelement as a function of a section over the centreline.

For both those cases you will have many/no section properties why having a single section on the element wont make any sense.

Hope the above makes sense

IsakNaslundBh:
Closing this as implementation PRs have now been merged. 
https://github.com/BHoM/BHoM/pull/458
https://github.com/BHoM/BHoM/pull/479
https://github.com/BHoM/BHoM/pull/481

PR #456, opened 2019-04-14
Geometry_oM #453 add initial solid geometry representations

### LINKED ISSUE #453: Geometry_oM : Add BoundaryRepresentation class for solid objects
To ensure a valid solid representation, plan for this class to be create as immutable consisting of a ReadOnlyCollection of ISurfaces. 
In the immediate term therefore - no direct Create engine methods from Surfaces will be created - with only a series of Converts from other known valid and watertight representations. i.e. Software converts or primitive solid objects

### LINKED ISSUE #454: Geometry_oM : Add ISolid Interface

### LINKED ISSUE #455: Geometry_oM : Add basic primitive solid representations
Starting with:
- [x] Sphere
- [x] Cylinder 
- [x] Cone
- [x] Cuboid

al-fisher:

Closes #453
Closes #454
Closes #455 

Creation of initial Solid Geometry representations in the BH.oM.Geometry namespace. As mentioned in the referenced Issues - this will for the immediate term be for representation in BHoM and translation between external libraries and platforms. Beyond Converts, Engine methods will necessarily be kept to a minimum.

### DIFF
--- a/Geometry_oM/Geometry_oM.csproj
+++ b/Geometry_oM/Geometry_oM.csproj
@@ -68,6 +68,7 @@
     <Compile Include="Solid\Cylinder.cs" />
     <Compile Include="Solid\ISolid.cs" />
     <Compile Include="Solid\Sphere.cs" />
+    <Compile Include="Solid\Torus.cs" />
     <Compile Include="Surface\Extrusion.cs" />
     <Compile Include="Surface\ISurface.cs" />
     <Compile Include="Surface\Loft.cs" />
--- a/Geometry_oM/Geometry_oM.csproj
+++ b/Geometry_oM/Geometry_oM.csproj
@@ -62,6 +62,12 @@
     <Compile Include="Misc\Tolerance.cs" />
     <Compile Include="Misc\TransformMatrix.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Solid\BoundaryRepresentation.cs" />
+    <Compile Include="Solid\Cone.cs" />
+    <Compile Include="Solid\Cuboid.cs" />
+    <Compile Include="Solid\Cylinder.cs" />
+    <Compile Include="Solid\ISolid.cs" />
+    <Compile Include="Solid\Sphere.cs" />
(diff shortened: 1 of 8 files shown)


pawelbaran:
Looking good. One question, is `BoundaryRepresentation` just a watertight `PolySurface` or are there further differences? Plus, (looping @FraserGreenroyd in) will Solid be equivalent to a Volume? Finally, shouldn't we create `IElement3D` straight away? When I think about it, I can imagine creating a 3D laundry in a few-days sprint!

BTW, let's start raising issues for the basic methods that are missing - maybe @LMarkowski could make some first contributions.

al-fisher:
> One question, is `BoundaryRepresentation` just a watertight `PolySurface` or are there further differences?

Essentially - yes.
This is needed to enable clear translation and interop with platforms that can support solid geometry as Boundary Representations. I think it is useful to preserve a clear distinction between Solid (Immutable) `BoundaryRepresentation`s and a simple `PolySurface`s.
A Solid will naturally have a Queriable Volume. For a generalised `BoundaryRepresentation` this is non-trivial so have not included in this MVP. Further discussion needed - but have considered adding a Readonly Volume parameter or future CustomData/Fragment which could be set at Conversion if needed. One for later.

al-fisher:
@LMarkowski @pawelbaran 

> Let's start raising issues for the basic methods that are missing 

👍 Please do.
There will be some complications for some of the software converts - I will also raise some issues for discussion. Cheers

pawelbaran:
I am not sure about the BRep abbreviation. Most of the BHoM users are grown on GH - this might be confusing for them to see that their good old Brep needs to be closed :V (similar issue to GH Plane vs BHoM Plane/CoordinateSystem)

al-fisher:
Thanks both @pawelbaran @epignatelli - as a general rule naturally trying to minimise abbreviations in the Object Model, so went with full `BoundaryRepresentation`. Following on from your point also @pawelbaran suggest keeping full verbose type name to also more explicitly state the clear difference between a `BH.oM.Geometry.BoundaryRepresention` vs `Rhino.Geometry.Brep`

PR #496, opened 2019-06-06
Reflection_oM: DoNotExposeAttribute

### LINKED ISSUE #439: Reflection_oM: Add attribute to not expose a method in the UI
We usually aim at transparency between what we code in visual studio and all the available components in the UI, i.e. every method that is coded in the Engine is exposed as a component in the UI.

This connection can be broken by making the method `private`, but this is not always a convenient choice, since it affects its accessibility at developer level too.

To get the best from both of them, we resort to `Attribute`s. An attribute like `[NotExposed]` would be a good trade off to have a method implemented but not exposed to the UI.

epignatelli:

Closes #439 

Create the `DoNotExpose` attribute to allow flagging methods, constructor, fields and properties that we do not want to expose to the UIs.
For method it is pretty intuitive. The constructor target is for things like `Query` objects, or any non-standard object that has a constructor we want to obfuscate. For fields and properties, it can be used together with the `Explode` component, to want what we may want to hide form the UI.

For the name, I started with `UIHiddenAttribute`, moved to `NotExposedAttribute` and settled to `DoNotExpose` inasmuch as it expresses the action (as a `delegate`) passed to the UI logic.

Happy to have thoughts about the above!

### DIFF
--- a/Reflection_oM/Attributes/UIHiddenAttribute.cs
+++ b/Reflection_oM/Attributes/UIHiddenAttribute.cs
@@ -1,0 +1,52 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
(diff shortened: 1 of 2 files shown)


epignatelli:
sorry @IsakNaslundBh I have just see your card. Closing this.

adecler:
@epignatelli, can the branch be deleted then?

epignatelli:
Yea, sorry, I'll be more careful.

PR #498, opened 2019-06-07
Issue497-BoundindBoxOperatorNullCheck

### LINKED ISSUE #497: Geometry_oM: missing null check in BoundingBox operators

epignatelli:

Closes #497

The UI breaks when the operator is called with null arguments. I will treat it in the UI as well, but even if this may be a little forced, I think it is safer to add null checks here.


epignatelli:
Thanks @pawelbaran , agreed.
We shouldn't have null checks at this level of the pyramid, I added here an extra guard, since the UI is constantly throwing exceptions when no geometry can be displayed.

PR #508, opened 2019-06-18
BHoM: Fragments

### LINKED ISSUE #476: BHoM: Add IFragment interface
`IFragment` to be added as base interface for entity component system

Done as first step towards solving: https://github.com/BHoM/BHoM/issues/293

FraserGreenroyd:
Fixes #476 

### Test files
N/A

#### Added
 - Added Interface for Fragments to Base oM
 - Added FragmentProperties to Base oM object
#### Removed
 - Removed Fragment prototype from Environment objects to avoid clashes with base objects

### DIFF
--- a/BHoM/BHoMObject.cs
+++ b/BHoM/BHoMObject.cs
@@ -35,6 +35,8 @@ namespace BH.oM.Base
 
         public string Name { get; set; } = "";
 
+        public List<IBHoMFragment> FragmentProperties { get; set; } = new List<IBHoMFragment>();
+
         public HashSet<string> Tags { get; set; } = new HashSet<string>();
 
         public Dictionary<string, object> CustomData { get; set; } = new Dictionary<string, object>();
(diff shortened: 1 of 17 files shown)


BHoM/BHoMObject.cs
```diff
@@ -35,6 +35,8 @@ public class BHoMObject : IBHoMObject
 
         public string Name { get; set; } = "";
 
+        public List<IBHoMFragment> FragmentProperties { get; set; } = new List<IBHoMFragment>();
```
adecler: Since this is something we are going to have to use a lot and for a long time, I am going to be a bit picky here: I would prefer this property to be named `Fragments` instead of `FragmentProperties`. Makes it nicer in the UI and easier on the tongue :-)

rwemay: Agree!

FraserGreenroyd: Hehe I was trying to get away without a large trickle update - but will get it sorted today :smile:

FraserGreenroyd: All changed (as far as I can find)

adecler:
Looking good on my side 👍 
Will let someone that can do more in-depth testing approve though.

PR #509, opened 2019-06-19
DataManipulation_oM: rename `*Query` to `*Request`

### LINKED ISSUE #499: DataManipulation_oM : Rename Queries to Requests
`IQuery` and `Query` objects need to be renamed to avoid confusion with the `Engine.*.Query` class. Agreed proposal was to rename as `Request`

epignatelli:

Closes #499 

To summarise the changes:
- `BH.oM.DataManipulation.Queries.IQuery` into `BH.oM.Data.Requests.IRequest`
- `BH.oM.DataManipulation.Queries.BatchQuery` into `BH.oM.Data.Requests.BatchRequest`
- `BH.oM.DataManipulation.Queries.FilterQuery` into `BH.oM.Data.Requests.FilterRequest`
- `BH.oM.DataManipulation.Queries.CustomQuery` into `BH.oM.Data.Requests.CustomRequest`

They will also be moved from the `DataManipulation_oM` to a new project `Data_oM`.
The `DataStructure_oM` will also be moved to the new `Data_oM` project.

An example of transition in the actual code is:
`public override IEnumerable<object> Pull(IQuery query, Dictionary<string, object> config = null)` to
`public override IEnumerable<object> Pull(IRequest query, Dictionary<string, object> config = null)`

or 

`m_linkIn.SendData(new List<object>() { PackageType.UpdateProperty, new Tuple<FilterQuery,string,object>(filter,property, newValue), config, RevitSettings });` to 
`m_linkIn.SendData(new List<object>() { PackageType.UpdateProperty, new Tuple<FilterRequest,string,object>(filter,property, newValue), config, RevitSettings });`

Would you mind referencing (linking) this pr from any child pr that adapts to these changes?
This way we can keep track of what's been raised already.


Data_oM/Structures/GraphLink.cs
```diff
@@ -20,7 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
-namespace BH.oM.DataStructure
+namespace BH.oM.Data.Structures
```
IsakNaslundBh: Not wanting to be a pain, but is there a slight risk that the `Structures` here could be confused by the general `Structure` namespace and discipline?

Could assume that this would be a namespace for data related to the discipline.

al-fisher: I am actually less worried about this. To some extent we need to continue to rely on context as the English language has a habit of repetition of words.
That said - one option could be to rename `Data.Structures` as `Data.Elements` giving consistency across other namespaces?

FraserGreenroyd:
As discussed with @al-fisher , I have aligned this PR with the fragments PR which was missing from this branch and impacting the ability to test Query -> Request changes in Environmental toolkits :smile:

Data_oM/Structures/GraphLink.cs
adecler: `Elements` doesn't make much sense here. 
Why not use the same term as Microsoft uses for C#: `Collections`.

epignatelli: > Looking more deeply at the code - thinking to suggest renaming `Data.Structures` to `Data.Elements` for the reasons stated above. Do think it might be clearer giving parity across other namespaces.
> The meaning of "Elements" and "Structures" are interchangeable here really so could help?
> What do we think?
> @adecler @epignatelli @rwemay interested in thoughts

If the aim of this is clarity, I think that `Data.Elements` is less clear and intuitive.
If the aim is consistency across namespaces, `Data.Elements` is definitely the way.

In a trade off between the two, I would lean towards clarity (i.e the `Data.Structures`), as it is easier to use and the risk of confusion with the `Structure` namespace is very low.
I also think that consistency underlines clarity as objective, so another point for me in favour of `Data.Structures`

adecler:
@al-fisher. See my comment in code review from Isak.

Data_oM/Structures/GraphLink.cs
al-fisher: `Collections` I like.

epignatelli: If it's decided I can act on it

al-fisher: Agree with points above about consistency vs clarity. 
`Elements` was not really helping there.
For me `Collections` is actually the clearest of the lot.

al-fisher:
> @al-fisher. See my comment in code review from Isak.

Perfect! Thanks - think all in good agreement on that one!
@epignatelli are you have to action tweaking?

Cheers

JosefTaylor:
Nothing needed from SAP2000, as I never reference the DataManipulation om or override the Pull. Shall I close my issue and remove my line from the checklist?

epignatelli:
I think the problem lies in the SAP2000_Test project, which is referencing the DataManipulation_oM.

Data_oM/Requests/BatchRequest.cs
```diff
@@ -21,17 +21,16 @@
...
 {
-    public class BatchQuery : IQuery
+    public class BatchRequest : IRequest
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        public List<IQuery> Queries { get; set; } = new List<IQuery>();
+        public List<IRequest> Queries { get; set; } = new List<IRequest>();
```
al-fisher: Would suggest Property `Queries` should now be renamed `Requests`

Data_oM/Requests/CustomRequest.cs
```diff
@@ -20,9 +20,9 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
-namespace BH.oM.DataManipulation.Queries
+namespace BH.oM.Data.Requests
 {
-    public class CustomQuery : IQuery
+    public class CustomRequest : IRequest
```
al-fisher: Same with the CustomRequest Property `Query` to be renamed. To `Request` or `RequestText`.

PR #528, opened 2019-07-19
Diffing_oM initial implementation

### LINKED ISSUE #506: Diffing_oM: Implementation
Implement Diffing_oM. This should include a `Delta` class with the [schema agreed in AECDeltas](https://github.com/aecdeltas/aec-deltas-spec/wiki/Delta-Specification):

```
+ ToCreate {get; private set;} // or "new"
+ ToDelete {get; private set;} // or "old"
+ ToUpdate {get; private set;} // or "modified"
+ ProjectId {get; private set;}
+ Version {get; private set;}
+ Timestamp {get; private set;}
+ Signature {get; private set;}
+ Sender {get; private set;}
```

### Other references
https://github.com/BHoM/BHoM_Engine/issues/1031

alelom:
Closes #506 

Any other change needed for a first implementation of the Diffing will be appended here.

### Changelog
* Added Diffing_oM project
* Added `BH.oM.Diffing.Delta` class


FraserGreenroyd:
As discussed in BHoM call: @epignatelli and @al-fisher and @pawelbaran to review this draft PR to provide initial feedback to @alelom in conjunction with https://github.com/BHoM/BHoM_Engine/pull/1103

Diffing_oM/DiffingHashFragment.cs
```diff
@@ -0,0 +1,87 @@
...
+    public class DiffHashFragment : IBHoMFragment
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public string Hash { get; private set; }
+        public string PreviousHash { get; private set; } = null;
+
+        public DiffProjFragment DiffingProject { get; private set; }
```
epignatelli: I think hash should be nothing more than a simple `string`. Agree that `PreviousHash` could be useful, maybe `ParentHash` is more descriptive?

Although, I find this is not the best place to store the `DiffProjFragment`. Hash should be as lightweight as possible, to minimise the `Delta` payload.

Diffing_oM/DiffingHashFragment.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /***************************************************/
+
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public DiffHashFragment(string hash, DiffProjFragment diffingProject)
```
epignatelli: Same as above for constructors, they should be a `Create` method in the `Engine`.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        public List<IBHoMObject> ToUpdate { get; private set; }
+        public List<string> ToUpdate_hashes { get; private set; }
+
+        public List<IBHoMObject> Unchanged { get; private set; }
+        public List<string> Unchanged_hashes { get; private set; }
+
+        [Description("The dict key is the modified object hash; tuple.item1 is a list of all modified props names; tuple.item2 the list of new values.")]
+        public Dictionary<string, Tuple<List<string>, List<string>>> ModifiedPropsPerObject { get; private set; }
+
+        public DiffProjFragment DiffingProject { get; private set; }
```
epignatelli: I don't see the need of encapsulating the `ProjectId` and `Revision` information into an object, because they do not have a particular logic that requires a `class`. 

Maybe it is better to replace `DiffProjFragment` with the two properties straight away?

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        public List<IBHoMObject> ToCreate { get; private set; }
+        public List<string> ToCreate_hashes { get; private set; }
+
+        public List<IBHoMObject> ToDelete { get; private set; }
+        public List<string> ToDelete_hashes { get; private set; }
+
+        public List<IBHoMObject> ToUpdate { get; private set; }
+        public List<string> ToUpdate_hashes { get; private set; }
+
+        public List<IBHoMObject> Unchanged { get; private set; }
```
epignatelli: Why do we have `Unchanged` objects as well? 
This is going to be very very heavy if we diff a big model.

I am not sure why it has been included, the only way I can think about this is for a first "push". But at that point all the objects would be `ToCreate`.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public Delta(DiffProjFragment diffingProject, List<IBHoMObject> toCreate, List<IBHoMObject> toDelete, List<IBHoMObject> toUpdate, List<IBHoMObject> unchanged)
+            : this(diffingProject, toCreate)
+        {
+            ToDelete = toDelete;
+            ToUpdate = toUpdate;
+            Unchanged = unchanged;
+
+            if (ToDelete_hashes == null && toDelete != null)
+                ToDelete_hashes = GetHashes(toDelete);
```
epignatelli: if the condition is not satisfied what happens? 
Does the `Delta` fail?
Same for the next ones.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        /***************************************************/
+
+
+        /***************************************************/
+        /**** Private methods                           ****/
+        /***************************************************/
+
+        private List<string> GetHashes(List<IBHoMObject> objs)
+        {
+            return objs.Select(obj => obj.Fragments.OfType<DiffHashFragment>().First().Hash).ToList();
```
epignatelli: Watch out here, `obj.Fragments.OfType<DiffHashFragment>()` might return an empty list.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            if (ToDelete_hashes == null && toDelete != null)
+                ToDelete_hashes = GetHashes(toDelete);
+            if (ToUpdate_hashes == null && ToUpdate != null)
+                ToUpdate_hashes = GetHashes(toUpdate);
+            if (Unchanged_hashes == null && Unchanged != null)
+                Unchanged_hashes = GetHashes(Unchanged);
+        }
+
+        public Delta(DiffProjFragment diffingProject, List<IBHoMObject> toCreate, List<string> toCreate_hashes, List<IBHoMObject> toDelete, List<string> toDelete_hashes, List<IBHoMObject> toUpdate, List<string> toUpdate_hashes, List<IBHoMObject> unchanged, List<string> unchanged_hashes)
```
epignatelli: Uhm this one is tricky. I spotted it when reviewing the engine. 
I think it is a bad idea to expose this publicly, because it is very surgical: you need to pass the objects, calculate the hashes, and make sure they are in the same position in the list?

A couple points:
- `ToCreate` doesn't require hashes, it just requires the objects. I am not sure why the hashes would be useful to the receiver.
- `ToDelete` only requires the hashes. That's how the receiver will filter out the objects to delete.
- `ToUpdate` would require both, but I think the hash should be stored in the object itself.
- `Unchanged` is not giving any more info than the rest, I'd remove it altogether.

The signature then would be:
```c#
public Delta(string projectName, List<IBHoMObject> toCreate, List<string> toDelete, List<IBHoMObject> toUpdate)
```

Diffing_oM/Diffing_oM.csproj
```diff
@@ -0,0 +1,62 @@
...
+    <RootNamespace>Diffing_oM</RootNamespace>
+    <AssemblyName>Diffing_oM</AssemblyName>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>..\Build\</OutputPath>
```
epignatelli: Shall we also set the `Default namespace` property to `BH.oM.Diffing`?

Diffing_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Diffing_oM")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("Diffing_oM")]
+[assembly: AssemblyCopyright("Copyright ©  2019")]
```
epignatelli: `[assembly: AssemblyCopyright("Copyright © https://github.com/BHoM")]`

Diffing_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Diffing_oM")]
+[assembly: AssemblyDescription("")]
```
epignatelli: `[assembly: AssemblyDescription("BHoM Diffing oM")]`

Diffing_oM/Delta.cs
epignatelli: Also, if we give a default value to `ToDelete` and`ToUpdate`, this method and the one above can be easily merged, making all more consistent.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+namespace BH.oM.Diffing
+{
+    public class Delta : IObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public List<IBHoMObject> ToCreate { get; private set; }
+        public List<string> ToCreate_hashes { get; private set; }
```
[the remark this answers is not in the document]
adecler: Not sure about the `_hashes` properties either. I could go behind a single central `Hash` property though

Diffing_oM/Delta.cs
adecler: +1

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        public List<string> ToDelete_hashes { get; private set; }
+
+        public List<IBHoMObject> ToUpdate { get; private set; }
+        public List<string> ToUpdate_hashes { get; private set; }
+
+        public List<IBHoMObject> Unchanged { get; private set; }
+        public List<string> Unchanged_hashes { get; private set; }
+
+        [Description("The dict key is the modified object hash; tuple.item1 is a list of all modified props names; tuple.item2 the list of new values.")]
+        public Dictionary<string, Tuple<List<string>, List<string>>> ModifiedPropsPerObject { get; private set; }
```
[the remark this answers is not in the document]
adecler: Shouldn't it be `Dictionary<string, Dictionary<string, object>>` or am I missing something?

Diffing_oM/Delta.cs
adecler: +1. 
There might be a need to group a bunch of properties later but that can be handled then. Two properties, especially those two, feel like there are better off being exposed directly.

Also, I don't like the word fragment here as it clashes with the concept of fragments we already have. 
Edit: Ha, I see those classes are actual fragments, we have a Fragment property to store them. If you need to enforce their existence as properties, then maybe they shouldn't be fragments in the first place.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        public long Timestamp { get; private set; }
+        public string Author { get; private set; }
+
+        /***************************************************/
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public Delta(DiffProjFragment diffingProject, List<IBHoMObject> toCreate)
```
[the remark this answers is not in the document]
adecler: very important, that one

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+    public class Delta : IObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public List<IBHoMObject> ToCreate { get; private set; }
+        public List<string> ToCreate_hashes { get; private set; }
+
+        public List<IBHoMObject> ToDelete { get; private set; }
```
adecler: Shouldn't it just be a list of Guid? No need to transfer the whole object over, right?

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        public List<string> ToDelete_hashes { get; private set; }
+
+        public List<IBHoMObject> ToUpdate { get; private set; }
```
adecler: You have the `ModifiedPropsPerObject` property already, why do you need this?

Personally, I don't like the name `ModifiedPropsPerObject` and would simply have 
`Dictionary<string, Dictionary<string, object>> Updated` instead.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+            DiffingProject.Revision += 1;
+
+            Timestamp = DateTime.UtcNow.Ticks;
+            Author = Environment.UserDomainName + "/" + Environment.UserName;
+
+            if (ToCreate_hashes == null)
+                ToCreate_hashes = GetHashes(toCreate);
+        }
+
+        public Delta(DiffProjFragment diffingProject, List<IBHoMObject> toCreate, List<IBHoMObject> toDelete, List<IBHoMObject> toUpdate, List<IBHoMObject> unchanged)
```
adecler: Even if they were in the Engine.Create, I am not sure I would like those constructors.

Why not have something like 
`Engine.Compute.Delta(List<IBHoMObject> oldObjects, List<IBHoMObject> newObjects,...)`

and no direct method in Engine.Create to produce a delta as I don't think we need any?

Diffing_oM/DiffingHashFragment.cs
adecler: I am not sure this class should exist. I don't understand why` DiffingProject` is in there so that leaves only two properties... not sure this is worth isolating in a separate class.

Diffing_oM/DiffingProjectFragment.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace BH.oM.Diffing
+{
+    public class DiffProjFragment : IBHoMFragment
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public string ProjectName { get; private set; }
+        public string ProjectId { get; private set; }
```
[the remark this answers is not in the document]
adecler: +1

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: This is due to the structure of the output of the [Engine.Testing.Query.IsEqual](https://github.com/BHoM/BHoM_Engine/blob/master/Testing_Engine/Query/IsEqual.cs#L50) method.
I agree with @adecler choice -- my plan was to implement it like that anyway.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace BH.oM.Diffing
+{
+    public class Delta : IObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public List<IBHoMObject> ToCreate { get; private set; }
```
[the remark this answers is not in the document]
alelom: Our idea was that the Delta is immutable, right? Then I should probably implement IImmutable.

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: As I said above, our idea was that the Delta is immutable, right? Then I should probably implement IImmutable. What do you think?

Diffing_oM/DiffingHashFragment.cs
alelom: Originally I separated DiffingProject and Hash fragments during WIP because the DiffingProject is required to have a unique Hash. I thought one might want to create the diffing project independently and then use it to create hashes. Might as well merge them now.

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
al-fisher: Yes - `Delta` should indeed implement the `IImmutable` interface.

Diffing_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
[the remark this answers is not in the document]
alelom: done

Diffing_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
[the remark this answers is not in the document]
alelom: done

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
al-fisher: Yes - `Delta` should indeed implement the `IImmutable` interface.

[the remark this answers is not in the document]
alelom: I've implemented the `IImmutable` interface.

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: I've implemented the `IImmutable` interface.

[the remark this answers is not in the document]
epignatelli: I think it still needs a little bit of work that one.

I've put together a flash wiki on how to implement `IImmutable`:
https://github.com/BHoM/documentation/wiki/The-IImmutable-interface

Le me know if that helps!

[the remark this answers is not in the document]
alelom: Thank you, absolutely helpful! :)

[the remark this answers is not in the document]
alelom: I've implemented the changes following your wiki, however, before removing the additional constructors (which I agree to be the right thing) I think we need to review a bit how Fragments work. 
The current arrangement optimises the algorithm in that the hashes are stored as well, and you don't need an additional lookup among the objects' fragments (see [`GetHashFragment()`](https://github.com/BHoM/BHoM_Engine/blob/56e9d72a67018fcb33193eb1289040c0c99ff540/Diffing_Engine/Query/GetHash.cs#L13-L30)). This can be solved when [our conversation around the fragments is done](https://github.com/BHoM/BHoM/issues/538).

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: While I agree with you both, the reason why there are the hashes properties is for performance, and it's explained [here](https://github.com/BHoM/BHoM/issues/538). In fact, I am using a [`GetHashFragment()` method](https://github.com/BHoM/BHoM_Engine/blob/56e9d72a67018fcb33193eb1289040c0c99ff540/Diffing_Engine/Query/GetHash.cs#L13-L30) just to retrieve the hashes of the objects, and that has a computationl cost that I wanted to avoid if possible. 
This is due to the fact that we are forced to use a `List<Fragment>` instead of `CustomData`. 
I agree this is not an optimal solution and I will be happy to remove these properties when we reach an agreement on [how the fragments should be stored](https://github.com/BHoM/BHoM/issues/538).

Diffing_oM/Delta.cs
alelom: The reason why the `HashFragment` and the `DiffProjFragment` are distinct is because I designed the diffing so that one can define a specific Project for which he can have multiple HashFragment defined (e.g. if a different criteria is needed in the hash generation under the same Project).

While I'm not sure I agree with your proposed solution, since it is true we still don't have an use case for my proposal, I will do as you say.

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: I'll resolve the conversation for now, with the `_hashes` properties to be removed in a next iteration after we will have changed how the fragments are stored.

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: I've modified as Arnaud proposed, except I added also the `object` property that represents the old value of the property. Final form is:
`Dictionary<string, Dictionary<string, Tuple<object,object>>>`
where tuple.Value1 is value in "SetA" (or current, or new), tuple.Value2 is the value in "setB" (or old, or modified).

Diffing_oM/Delta.cs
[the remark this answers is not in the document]
alelom: For now anyway I'll remove all the _hashes properties.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,124 @@
...
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public Delta(DiffProjFragment diffingProject, List<IBHoMObject> toCreate)
+        {
+            ToCreate = toCreate;
+
+            DiffingProject = diffingProject;
+            DiffingProject.Revision += 1;
```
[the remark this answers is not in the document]
alelom: Added null check, thanks!

Diffing_oM/DiffingHashFragment.cs
alelom: Implemented IImmutable.

Diffing_oM/Delta.cs
alelom: I will remove it.

Diffing_oM/Delta.cs
alelom: Removed.

Diffing_oM/Delta.cs
alelom: Added Count check.

Diffing_oM/Delta.cs
alelom: Removed all hashes properties and  the Unchanged.

Diffing_oM/Properties/AssemblyInfo.cs
alelom: Done

Diffing_oM/Properties/AssemblyInfo.cs
alelom: Done

Diffing_oM/Delta.cs
alelom: Done.

Diffing_oM/Delta.cs
alelom: That's right. At the moment of trasfer we can just extract the hashes from the objects. Keeping the objects it's useful when using the Diffing outside the context of data transfer -- e.g. when exploding the Delta in the UI to see the differences.

Diffing_oM/Delta.cs
alelom: Same as above (https://github.com/BHoM/BHoM/pull/528#discussion_r315252332)

Diffing_oM/Delta.cs
alelom: Now implemented the IImmutable and kept only one constructor.

Diffing_oM/Diffing_oM.csproj
alelom: added `<RootNamespace>BH.oM.Diffing</RootNamespace>`

epignatelli:
Removing milestone during Sprint Closure call as this is parked

alelom:
Closing as been superseded by https://github.com/BHoM/BHoM/pull/541

PR #541, opened 2019-08-28
BHoM: Diffing_oM initial implementation 

### LINKED ISSUE #506: Diffing_oM: Implementation

alelom:
Closes #506 

Ported https://github.com/BHoM/BHoM/pull/528 and adapted to the newly agreed concept as discussed with @al-fisher 

### Changelog
* Added Diffing_oM project
* Added `BH.oM.Diffing.Delta` class


Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace BH.oM.Diffing
+{
+    public class Delta : IObject, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public List<IBHoMObject> NewObjs { get; }
```
adecler: Do those need to be `IBHoMObject` or can they be `IObject` ?

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public List<IBHoMObject> OldObjs { get; }
+
+        public List<IBHoMObject> Modified { get; }
```
adecler: A few things here:
- I don't like the shortening of variable names. It reduces readability for no clear benefit
- To be consistent with the other two, should `Modified` be called `ModifiedObjects` ?
- The information stored in `ModifiedObjects` and `ModifiedPropsPerObject` is overlaping, right? So we should only keep one of the two.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using BH.oM.Reflection.Attributes;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Diffing
+{
+    public class Delta : IObject, IImmutable
```
adecler: If we sort out the comment below on unnecessary overlapping variables, the object doesn't have to be `IImmutable`, right? The only reason why the Delta would need to be immutable is if it was containing a hash property (for the whole object). I think it probably should, but if not, immutability looses it purpose.

Edit: I suppose you are storing the Hash in a fragment instead of as a property. In that case, this highlights the fact that our fragment system is still incomplete at this stage: We need a `RequiredFragments` attribute for this class. Otherwise, there is no garanty that the hash fragment will be there. @al-fisher , @FraserGreenroyd , FYI.

Diffing_oM/Delta.cs
```diff
@@ -0,0 +1,72 @@
...
+        /***************************************************/
+
+        [Description("Creates new Delta object with information on the new, old or modified object, and (if exists) the Diffing Stream that contains them.")]
+        [Input("onlySetA", "Objects existing exclusively in the 'primary' set, e.g. the 'new' objects.")]
+        [Input("onlySetB", "Objects existing exclusively in the 'secondary' set, e.g. the 'old' objects.")]
+        [Input("modified", "Objects existing in both sets that have some differences in their properties.")]
+        [Input("unModified", "Objects existing in both sets that hold no differences in their properties.")]
+        [Input("modifiedPropsPerObject", "Dictionary holding the differences in properties of the 'modified' objects. See the corresponding property description for more info.")]
+        [Input("diffingStream", "If the Delta is the result of a diffing in the context of a Stream, this is the stream that holds the objects. Otherwise null.")]
+        public Delta(List<IBHoMObject> setA, List<IBHoMObject> setB, List<IBHoMObject> modified, Dictionary<string, Dictionary<string, Tuple<object, object>>> modifiedPropsPerObject = null)
```
adecler: any reason for the variable names `setA` and `setB` instead of 'newObjects' and 'oldObjects' ?

Diffing_oM/HashFragment.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace BH.oM.Diffing
+{
+    public class HashFragment : IBHoMFragment
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public string Hash { get; private set; }
+        public string PreviousHash { get; private set; } = null; // Until we solve the 'Persistent GUID', this is the current workaround.
```
adecler: I disagree with the private set for those two. This can only happen if the class is `IImmutable`. 
Making the class immutable here would not make me happy though 😝. A class can only be immutable if it has non-orthogonal properties, which is not the case here. Don't confuse the concept of `IImmutable` and the conceptual idea that you don't want your hash fragment to be modified.

Diffing_oM/HashFragment.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /***************************************************/
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public HashFragment(string hash, string previousHash = null)
```
adecler: Same thing, constructor shouldn't be here

Diffing_oM/HashFragment.cs
```diff
@@ -0,0 +1,78 @@
...
+            PreviousHash = previousHash;
+        }
+
+        /***************************************************/
+
+        /***************************************************/
+        /**** IBHoMFragment stuff (not needed)          ****/
+        /***************************************************/
+
+        public Guid BHoM_Guid { get; set; } = Guid.NewGuid();
```
adecler: This code should be deleted

Diffing_oM/Stream.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using BH.oM.Reflection.Attributes;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Diffing
+{
+    public class Stream : BHoMObject, IImmutable
```
adecler: Same thing, I don't see a real reason to make this immutable

Diffing_oM/Stream.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public string StreamId { get; } // Generally a GUID
+        public string StreamName { get; } // Like Github's Branch name
```
adecler: Can't the Name property be used for that? Otherwise, we will end up with two name inputs when creating a new Stream using the custom constructor.

Diffing_oM/Stream.cs
```diff
@@ -0,0 +1,51 @@
...
+        public string Revision { get; } // Perhaps this can become just a "revisionName" or "comment", like Github's Summary. After all, Revision can be given by Timestam+Author info
+        public long Timestamp { get; }
+        public string Author { get; }
+
+        public IEnumerable<IBHoMObject> Objects { get; }
```
adecler: Ok, that's me commenting before looking at the Engine part of the code but I think it is important to approach this with fresh eyes first...

Should it be `Deltas` instead of `Objects` here?

Diffing_oM/Stream.cs
```diff
@@ -0,0 +1,51 @@
...
+        /***************************************************/
+        /**** Constructor                               ****/
+        /***************************************************/
+
+        [Description("Creates new Diffing Stream")]
+        [Input("objects", "Objects to be included in the Stream")]
+        [Input("streamName", "If not specified, the name will be `UnnamedStream` followed by UTC")]
+        [Input("streamId", "If not specified, streamId will be a GUID.Revision is initally 0")]
+        [Input("revision", "If not specified, revision is initially set to 0")]
+        public Stream(IEnumerable<IBHoMObject> objects, string streamName = null, string streamId = null, string revision = null)
```
adecler: Should be moved to the engine

Diffing_oM/Delta.cs
alelom: `IBHoMObjects` at the moment.

Diffing_oM/Delta.cs
alelom: I'll keep `ModifiedPropsPerObject`

alelom: I've renamed the variables to avoid the shortening.
I agree both `ModifiedObjects` and `ModifiedPropsPerObject` are overlapping but we decided with Al to include them for ease of use in the UI: 
image

The idea is to then transfer the minimum amount of information (without repetition) when doing the push, for which we might decide to add another class `DeltaMessage`

Diffing_oM/Delta.cs
alelom: No, changed

Diffing_oM/Stream.cs
alelom: Might become such in a future iteration -- at the moment I see it working differently, see Engine

Diffing_oM/HashFragment.cs
alelom: I agree with the `private set` -- however I'd rather have this class as `Immutable` because that would be my intent.
I am not sure I'm getting your point on the use on `IImmutable` though. 
Are you saying that is our internal policy to implement `IImmutable` like that? It's not in its definition: [immutable types](https://en.wikipedia.org/wiki/Immutable_object) (objects or data structures) are types that, once initialized, cannot change their internal state -- no relation at all with their properties

Diffing_oM/Stream.cs
alelom: Removed streamName

Diffing_oM/Stream.cs
alelom: It contains objects that hold a Hash fragment whose immutability must be guaranteed over the lifetime of the Stream

Diffing_oM/Delta.cs
alelom: Agreed, I'd like to discuss this during our chat tomorrow @adecler

al-fisher: Just capturing the useful chat earlier. `RequiredFragments` could be an attribute on an Engine Method - stating this operation can only be performed on objects with certain Fragments attached.
On an object/property level however `RequiredFragments` as a concept should not exist. If the fragment properties are "required" they should be defining properties of the object. Objects must be definable and valid in absence of any fragments.

al-fisher: I do think this object should remain `Immutable`  
The properties:
`List<IBHoMObject> NewObjects` ,
`List<IBHoMObject> OldObjects`  &
`List<IBHoMObject> ModifiedObjects`
are not actually mutually perpendicular - they need to be a compatible state of lists defining the result of the diffing.  You can not (should not) interdependently modify one of these lists.

alelom: Thanks @al-fisher. I would like to add that regardless of the orthogonality I think that this class should be immutable. A Delta object should never be modified once created -- its existence/definition is tied to the diffing calculation in a specific moment in time and configuration.

al-fisher: Agreed.

epignatelli:
According to Sprint Closure call - @al-fisher , @alelom is asking for merging this to solve the milestone. To be taken offline.

al-fisher:
Thanks all - agreed. Previous offline discussion with @alelom, happy with the scope of this initial prototype - with further improvements and potential changes to be in future PRs.
Will provide final review - sorry for delay 😄

PR #560, opened 2019-09-30
Add embodied carbon to material fragments

### LINKED ISSUE #559: Structure_oM: add embodied carbon to materials
#### Description:
Add a property for embodied carbon, based on a very simplistic, but internally consistent life-cycle analysis.

JosefTaylor:

Closes #559 

Added embodied carbon to structural material fragments

### Changelog

Added property EmbodiedCarbon as a double for structural material fragments.

### Additional comments
There's an open discussion of where carbon accounting should be kept- This is intended to address carbon accounting at initial concept stages by structural engineers, with an aim of getting carbon savings into our design at the outset. More detailed carbon accounting methods are being developed by the environments team, which may supplement, but likely not supplant, this approach, thus it seems reasonable to implement it. There may also be utility in promoting it to a non discipline-specific namespace, however with our current use of MaterialFragments there would seem to be no place to promote it to.


JosefTaylor:
The way our derived properties compute, for most things, presents a problem, I think. If carbon is calculated for a bar as an engine query, it won't show up on explode. I've just mocked up an idea where carbon is calculated for a bar at creation, based on the sectionProp, material, and length, however I think this will not keep up if the bar is changed (nodes move).

Could the calculation be done as an engine query, but get called from the object definition, so that it is calculated whenever it is needed, i.e. on explode?

rwemay:
Interesting, I've thought about similar before: making derived properties from queries appear as properties in the UI, bar length being the simplest example for bar. I wonder if worth revisiting - might be mad/complex, but maybe there is a 'DerivedProperty' attribute or similar that we apply to a query method, that's called on Explode and GetProperty. There would obviously be no SetProperty, and the user can be warned about that if trying to set a property with same name as derived property. Might add a lot of overhead to some basic/very frequent methods like Explode/GetProperty though. @al-fisher , @adecler , is this mad?

IsakNaslundBh:
Amazing to get this in!!
Would not add this is a property to the bar for the reason you have already stated here @JosefTaylor .

As briefly discussed in the planning call yesterday and mentioned by @al-fisher , this might be an area where we would want to tweak the UI as you describe above. Rather than adding as a property, maybe the Explode always returns Embodied carbon for any AEC object, by calling the engine query method?

FraserGreenroyd:
@IsakNaslundBh @JosefTaylor this might be worth discussing in a wider context cause I can see this being useful to more than just structures oM as well, particularly Environment oM? Something to consider? :smile:

JosefTaylor:
OK @rwemay @IsakNaslundBh, I think at the moment I will take carbon out of bars, and instead create engine methods to get it. Then we can find a way to get it in the explode later.

FraserGreenroyd:
@JosefTaylor the approach makes sense to me, I'm just concerned over the duplication of code for an `EmbodiedCarbon` number and wonder if there is a better place to put it than the `Structure.oM` material fragment considering it will be relevant to pretty much all of the disciplines :smile:

adecler:
Sorry for jumping in at the 11th hour. If I understood correctly, you need properties that can be calculated directly from other properties of the object, right? If that's the case, you can do that already with something like 

```
public double CalculatedProperty { get { return A * B; } }
```

JosefTaylor:
@adecler the engine method is Bar.Mass() * Bar.SectionProperty.Material.EmbodiedCarbon;

Since BHoM doesn't reference the engine, I'd need to calculate the length of the bar from scratch, using the start and end nodes. Not hard, but also not sure if we want to get that much math into the oM. Thoughts?

Structure_oM/Elements/Bar.cs
```diff
@@ -60,6 +61,18 @@ public class Bar : BHoMObject, IElement1D, ILink<Node>
...
 
+        public double EmbodiedCarbon
+        {
+            get
+            {
+                Vector a = EndNode.Position - StartNode.Position;
+                return 
+                    Math.Pow((Math.Pow(a.X, 2) + Math.Pow(a.Y, 2) + Math.Pow(a.Z, 2)), 1 / 2)
+                    * SectionProperty.Area
+                    * SectionProperty.Material.EmbodiedCarbon;
```
FraserGreenroyd: From what I can see, the `SectionProperty` defaults to `null` - which means there is a risk of `Object reference not set to an instance of an object` if someone tries to obtain the `EmbodiedCarbon` without the `SectionProperty` being set. Obviously this can be resolved by checking if it is set before using it, but then we would be moving very much away from BHoM objects being simple property only objects. Obviously @adecler will be able to provide a final say on whether this would be acceptable within the framework, but I personally suspect we might be able to achieve what's been discussed in a different way that doesn't stray away from the BHoM philosophy for the oM :smile:

al-fisher: @FraserGreenroyd - you raise a good point on the `null` `SectionProperty`.

And also importantly @JosefTaylor - I do think you are right above about us needing to be careful about doing too much math~s~ 😄 in the oM itself.

Even if comfortable with this specific example - I am not sure how well this will scale. We will inevitably hit a property that will be _too_ complex to recreate outside of formal engine methods.

Propose we do keep computation of derived properties out of the oM entirely (unless perhaps implementing `IImmutable` Interface - different case).

Good to experiment and test limits! 👍 @JosefTaylor 

I would like us to explore @rwemay’s suggestion of a `DerivedProperty` `CustomAttribute` on a Query Method. 
Either that or a new discrete partial class for DerivedProperties in the engine.
Would prefer the former though as clearer.

This will, as said, require more framework/UI to give the desired exposure - however much better to add complexity there than in the oM I feel.

adecler: Agreed with all of that. I was mentioning that in-oM calculation just to make sure all our options are on the table. Given that it is already more complicated than A * B, I would not like it in the oM either. This feels like this is one that will require a decent amount of head scratching to get something that is not just an awkward addition.

JosefTaylor: shouldn't it be 'too many maths'?

JosefTaylor: I'll reverse that commit- It's not currently working anyway, and I agree with the sentiments above.

Once I do that, anything against going ahead and merging this in?

JosefTaylor:
@rwemay @IsakNaslundBh anything against merging this one?

IsakNaslundBh:
Only thing is that I think all the Datasets relating to material in any way will need to be updated (materials and sections).

Would like for that to be in order before merging this.

Issue #537, opened 2019-08-15
Should fragments be IObject instead of IBHoMObject?


alelom:
I had a go with the fragments during the implementation of the Diffing_Engine prototype.
When doing that, I had to store a `string` -- the hash -- in my objects, for which I duly created my `HashFragment`. Therefore, to store a string, I also had to store GUID, Name, and other properties I would never make use of.

The same is true for the "AdapterId". If we want to get rid of the CustomData, then we'll need an `AdapterIdFragment` to store only the id field, but that will still force all IBHoMObject properties to be there as well -- which will probably never be put to any use.

**What about having the Fragments as `IObject` instead?**

- that would not force on the coder many properties he might not need
- nothing stops the coder to make a fragment which implements `IBHoMObject` if needed

@al-fisher @adecler @epignatelli @FraserGreenroyd @IsakNaslundBh

adecler:
I am with you on this one. I don't see the point of having fragments as IBHoMObjects. 
On a thematic level, it doesn't make sense either: they are fragments of objects, not full BHoM objects.

al-fisher:
Yes - agreed too - sounds good.

alelom:
@al-fisher @adecler As agreed today I will action this in https://github.com/BHoM/BHoM/issues/563

PR #572, opened 2019-10-08
BHoM: Fragments to be `IObject` instead of `IBHoMObject`

### LINKED ISSUE #563: Fragments to be IObject instead of IBHoMObject
#### Broken rules:
See discussion in https://github.com/BHoM/BHoM/issues/537

Change Fragments to be IObject instead of IBHoMObject.

alelom:

Closes #563 

As agreed with @adecler, @al-fisher which asked me to action this (also, see discussion in https://github.com/BHoM/BHoM/issues/537):
changed IBHoMFragment to inherit IObject instead of IBHoMObject.

This allows to implement "lighter" fragments in all cases where the BHoMObject properties are not needed.

### Changelog

- Changed IBHoMFragment to implement IObject instead of IBHoMObject.

### DIFF
--- a/BHoM/Interface/IBHoMFragment.cs
+++ b/BHoM/Interface/IBHoMFragment.cs
@@ -22,7 +22,7 @@
 
 namespace BH.oM.Base
 {
-    public interface IBHoMFragment : IBHoMObject
+    public interface IBHoMFragment : IObject
     {
 
     }


IsakNaslundBh:
One thing to check before merging is if there are any toolkits using the fragments that have been relying on the fragments also implicitly giving the IBHoMObject implementation.

I would assume not, as if you do not also inherit from BHoMObject that would mean you would have to add CustomData, name, GetShallowClone etc in each of those classes, but would you be able to make a quick check @alelom ?

FraserGreenroyd:
> One thing to check before merging is if there are any toolkits using the fragments that have been relying on the fragments also implicitly giving the IBHoMObject implementation.
> 
> I would assume not, as if you do not also inherit from BHoMObject that would mean you would have to add CustomData, name, GetShallowClone etc in each of those classes, but would you be able to make a quick check @alelom ?

I've done that check as part of my review @IsakNaslundBh as that was my concern for Environment Fragments - I have found there are no issues on any Environmental toolkits as a result of this PR, but there is a need to remove the `BHoMObject` inheritance from Environmental Fragments which @alelom is raising an issue for me to resolve after this PR is merged :smile:

alelom:
Thanks both, perfect comment syncronisation 😄 

@FraserGreenroyd: thank you! Issue here https://github.com/BHoM/BHoM/issues/575
@IsakNaslundBh will do further checks

alelom:
From what I could gather, 
> One thing to check before merging is if there are any toolkits using the fragments that have been relying on the fragments also implicitly giving the IBHoMObject implementation.
> 
> I would assume not, as if you do not also inherit from BHoMObject that would mean you would have to add CustomData, name, GetShallowClone etc in each of those classes, but would you be able to make a quick check @alelom ?

this is never the case.

https://github.com/search?p=3&q=IBHoMFragment+org%3ABHoM+org%3ABuroHappoldEngineering&type=Code

alelom:
I had a go with the CodeAssistant and no issue in compiling with any repository.

alelom:
@al-fisher this is good for merge now.

PR #574, opened 2019-10-08
Fragments in BHoMObject now stored in a FragmentSet instead of in a List

### LINKED ISSUE #538: Fragment vs CustomData lookup cost: should we have a Dictionary<Fragment> (or HashSet<Fragment>) instead of List<Fragment>?
### Description 

Every time you need a specific fragment stored in an object, having a `List<Fragment>` forces you to **iterate the list and filter out that fragment**.

This means that in order to retrieve a specific type of fragment you need to use a method that does just that: iterate the list of fragments, identify the right type, return that.

This has an unavoidable cost in terms of performance which is sensibly higher than the cost of a Dictionary lookup, [which by definition is close to O(1)](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?redirectedfrom=MSDN&view=netframework-4.8#remarks). In this sense, `CustomData` is much more efficient, especially if you just need to store simple data (e.g. just a `string`, like with AdapterId)

### Practical example case
For example, in the Diffing_Engine I had to implement this [`GetHashFragment()`](https://github.com/BHoM/BHoM_Engine/blob/56e9d72a67018fcb33193eb1289040c0c99ff540/Diffing_Engine/Query/GetHash.cs#L13-L30) method just to retrieve the a specific `HashFragment` (and also, to make sure it was the only of its type among the other fragments).

After profiling the diffing, it's clear that [a high percentage of the computation is spent only looking for the HashFragment](https://github.com/BHoM/BHoM_Engine/issues/1119#cpu-profile).

### Proposal

After a chat with @al-fisher, we thought that instead of having a list of Fragments we might have a **Dictionary of fragments**, where the key is the fragment type and the value is the fragment object, such as:
```
Dictionary<Type, IBHoMFragment> Fragments
```

@al-fisher thought that in general we want to enforce that a certain type of fragment is not repeated in the Fragment list -- so this could also solve that.

Please comment with your thoughts.
@al-fisher @epignatelli @adecler @FraserGreenroyd @IsakNaslundBh

### LINKED ISSUE #573: Fragments to be `KeyedCollection<Type, IBHoMFragment>` instead of `List<IBHoMFragment>`
#### Description:

As discussed in https://github.com/BHoM/BHoM/issues/538

Implement Fragments as `KeyedCollection<Type, IBHoMFragment>` instead of `List<IBHoMFragment>`.

This will:
- immensely reduce the Lookup cost
- make them safer and behave as intended: no way to store the same Fragment twice

alelom:

Closes #573 
See also the discussion in #538 (Closes #538).

##### 1. Added the `FragmentSet` class that inherits from `KeyedCollection` to store the Fragments.
The FragmentSet class allows only `IBHoMFragment` types to be stored in the set (`where T : IBHoMFragment` ). 
The set therefore can include only one type of the same Fragment.

The FragmentSet has a method `GetKeyForItem` which returns the `Type` for a specific Fragment, that is the key of the collection. That is the only method that absolutely must be overridden, because without it the KeyedCollection cannot extract the keys from the items. See https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.keyedcollection-2?view=netframework-4.8

##### 2. Modified Fragments property to be `KeyedCollection<Type, IBHoMFragment>` instead of `List<IBHoMFragment>`

This has several effects:
- Compliance: Fragments now actually reflect their intended nature of an [Entity Component System](https://en.wikipedia.org/wiki/Entity_component_system)
- Safety/compliance: you cannot store more than one Fragment type per object, as originally intended
- Ease of use: no need for convoluted method to extract/append a specific Fragment to a BHoMObject
- Performance: lookup cost is lower as there is no need to check that the Fragment you are `get`ting is the only one in the collection for that Type. See also [profiling versus other collections](https://github.com/BHoM/BHoM/issues/538#issuecomment-521878391).

### Changelog

- Added the `FragmentSet` class that inherits from `KeyedCollection` to store the Fragments.
- Modified Fragments to be `KeyedCollection<Type, IBHoMFragment>` instead of `List<IBHoMFragment>` in IBHoMObject interface and BHoMObject class

### Additional notes
This change will require modifications in all Toolkits that implement Fragments.
I will open a separate issue for that. This one should be merged together with those changes.

### DIFF
--- a/BHoM/FragmentSet.cs
+++ b/BHoM/FragmentSet.cs
@@ -1,6 +1,6 @@
-/*
+﻿/*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
@@ -21,22 +21,15 @@
  */
 
 using System;
-using System.Collections.Generic;
+using System.Collections.ObjectModel;
 
 namespace BH.oM.Base
 {
-    public interface IBHoMObject : IObject
+    public class FragmentSet<T> : KeyedCollection<Type, T> where T : IBHoMFragment
     {
-        Guid BHoM_Guid { get; set; }
-
-        string Name { get; set; }
-
-        FragmentSet<IBHoMFragment> Fragments { get; set; }
-
-        HashSet<string> Tags { get; set; }
-
-        Dictionary<string, object> CustomData { get; set; }
-
-        IBHoMObject GetShallowClone(bool newGuid = false);
+        protected override Type GetKeyForItem(T item)
+        {
+            return item.GetType();
+        }
     }
 }
--- a/BHoM/FragmentSet.cs
+++ b/BHoM/FragmentSet.cs
@@ -25,9 +25,9 @@ using System.Collections.ObjectModel;
 
 namespace BH.oM.Base
 {
-    public class FragmentSet<T> : KeyedCollection<Type, T> where T : IBHoMFragment
+    public class FragmentSet : KeyedCollection<Type, IBHoMFragment>
     {
-        protected override Type GetKeyForItem(T item)
+        protected override Type GetKeyForItem(IBHoMFragment item)
         {
             return item.GetType();
         }
--- a/BHoM/FragmentSet.cs
+++ b/BHoM/FragmentSet.cs
@@ -40,12 +40,12 @@ namespace BH.oM.Base
 
         public bool AddOrReplace(IBHoMFragment fragment) // Slower than Add() or SetItem(), but easier to use.
         {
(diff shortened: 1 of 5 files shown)


BHoM/FragmentSet.cs
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.ObjectModel;
+
+namespace BH.oM.Base
+{
+    public class FragmentSet<T> : KeyedCollection<Type, T> where T : IBHoMFragment
```
adecler: As mentioned [here](https://github.com/BHoM/BHoM/issues/538#issuecomment-521878391), there is no need to have FragmentSet as a generic type. You can simply define it as `FragmentSet : KeyedCollection<Type, IBHoMFragment>`

alelom: Of course you're right, sorry! Changed now.

PR #582, opened 2019-10-16
Data_oM: Adding Dataset and Source object

### LINKED ISSUE #576: Data_oM: Add Dataset object
#### Description:

Need a `Dataset` class to work with all our serialised Datasets used by the Library_Engine.

Object to have properties:

- Source (this might need to be a separate object)
- Date of serialisation
- Data

Question is if this is best put in the Data_oM or if we should add a Library_oM for this. Important to not get this mixed up with any of the work done by @epignatelli 

@al-fisher @adecler @rwemay @FraserGreenroyd @JosefTaylor @enarhi @kayleighhoude @peterjamesnugent  any input welcome.

IsakNaslundBh:

Closes #576 

Adding Dataset object to be used with the serialised data in Datasets folder together with library engine. Added to formalise storing metadata such as time of creating and various soruce information

### Additional comments

@peterjamesnugent @kayleighhoude @JosefTaylor @enarhi @FraserGreenroyd

### DIFF
--- a/Data_oM/Library/Source.cs
+++ b/Data_oM/Library/Source.cs
@@ -35,9 +35,7 @@ namespace BH.oM.Data.Library
         /**** Properties                                ****/
         /***************************************************/
 
-        public string Hyperlink { get; set; } = "";
-
-        public string FilePath { get; set; } = "";
+        public string SourceLink { get; set; } = "";
 
         public string Title { get; set; } = "";
 
@@ -60,6 +58,8 @@ namespace BH.oM.Data.Library
         public string Country { get; set; } = "";
 
         public string Copyright { get; set; } = "";
+
+        public string Contributors { get; set; } = "";
 
         /***************************************************/
     }
--- a/Data_oM/Library/Source.cs
+++ b/Data_oM/Library/Source.cs
@@ -25,6 +25,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.ComponentModel;
 using BH.oM.Base;
 
 namespace BH.oM.Data.Library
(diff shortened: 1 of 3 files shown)


adecler:
Looks good to me. 
I just wonder if `HyperLink` and `FilePath` shouldn't be collapsed into one property.

al-fisher:
Thanks @IsakNaslundBh 
This is looking good. In addition to merging the `SourceLink` - wonder if we should do the same for 
`FigureReference`, `CodeReference` and `TableReference` - would all three be used simultaneously? or can we simply combine as a single field. 

Clarity could also be placed on each field by adding a `DiscriptionAttribute`

IsakNaslundBh:
Added first-pass descriptions to most of the properties. Any comments welcome.

Also not to sure of the Location name for country/region. If anyone has a better suggestion for that, it is more than welcome

Data_oM/Library/Source.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public string Schema { get; set; } = "";
+
+        [Description("Language of the source information")]
+        public string Language { get; set; } = "";
+
+        [Description("Country/Region of the source information")]
+        public string Location { get; set; } = "";
+
+        [Description("Any applicable copyright information assosiated with the data")]
```
peterjamesnugent: For consistency we should always refer to it as source information and not use variations of data, information etc.

Data_oM/Library/Source.cs
```diff
@@ -0,0 +1,73 @@
...
+        [Description("Title of the source. For example Book title, Paper title, Code title etc.")]
+        public string Title { get; set; } = "";
+
+        [Description("Author of the source of the information")]
+        public string Author { get; set; } = "";
+
+        [Description("Table, figure or code reference etc.")]
+        public string ItemReference { get; set; } = "";
+
+        [Description("Version of the source used")]
```
peterjamesnugent: Version of the source information used

Data_oM/Library/Source.cs
```diff
@@ -0,0 +1,73 @@
...
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Hyperlink or file path to where the source information used can be found")]
+        public string SourceLink { get; set; } = "";
+
+        [Description("Title of the source. For example Book title, Paper title, Code title etc.")]
+        public string Title { get; set; } = "";
+
+        [Description("Author of the source of the information")]
```
peterjamesnugent: Author of the source information

Data_oM/Library/Source.cs
```diff
@@ -0,0 +1,73 @@
...
+        [Description("Table, figure or code reference etc.")]
+        public string ItemReference { get; set; } = "";
+
+        [Description("Version of the source used")]
+        public string Version { get; set; } = "";
+
+        [Description("Publisher of the source information")]
+        public string Publisher { get; set; } = "";
+
+        public string Schema { get; set; } = "";
```
peterjamesnugent: There's no description here, are we expecting it to be a scheme reference such as name?

PR #586, opened 2019-10-17
DeepLearning_oM: implement most common artificial neural network schemas

### LINKED ISSUE #585: DeepLearning_oM: implement most common artificial neural network schemas
#### Description:

As per our last chat @al-fisher , going to implement the schemas in the BHoM

epignatelli:

Closes #585 

This pr agnostifies basic Deep Learning (DL) concepts to guarantee the interoperability between the most known DL libraries. You can check https://github.com/BHoM/TensorFlow_Toolkit, for an example.

### Test files

You can create any of the objects below by using the `CreateCustom` component in the UI and selecting the specific type you want.

### Changelog

- Add the `DeepLearning` namespace to the BHoM and some simple concepts: 
- Added neural network objects: `Graph` and `Sequential`
- Added most frequently used neural network operations:
  - Average Pooling 2d
  - Convolution 2d
  - Gated Recurrent Unit
  - Linear 
  - Long Short Term Memory
  - Max Pooling 2d
  - Max Unpooling 2d
  - Transposed Convolution 2d
- Added most frequently used activation functions:
  - LeaklyReLU
  - LogSoftmax
  - ReLU
  - Sigmoid
  - Softmax
  - Tanh
- Added most frequently used loss functions and their numerically stable equivalents when used with logits:
  - Binary Cross Entropy
  - Cross Entropy
  - L1
  - Mean Square Error
  - Negative Log Likelihood
- Added most frequently used optimisers:
  - Adam
  - Stochastic Gradient Descent

### DIFF
--- a/DeepLearning_oM/Layers/AvgPooling2d.cs
+++ b/DeepLearning_oM/Layers/AvgPooling2d.cs
@@ -1,0 +1,42 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using System;
+
+namespace BH.oM.DeepLearning.Layers
+{
+    public class AvgPooling2d : BHoMObject, IModule
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public Tuple<int, int> KernelSize { get; set; } = new Tuple<int, int>(2, 2);
+
+        public Tuple<int, int> Stride { get; set; } = new Tuple<int, int>(2, 2);
+
+        public Tuple<int, int> Padding { get; set; } = new Tuple<int, int>(0, 0);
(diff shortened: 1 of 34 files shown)


DeepLearning_oM/Losses/BCEWithSigmoid.cs
```diff
@@ -0,0 +1,28 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.DeepLearning.Losses
+{
+    public class BCEWithSigmoid : BinaryCrossEntropy
```
adecler: Not too keen on deeper inheritance chains

DeepLearning_oM/Layers/AvgPooling2d.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace BH.oM.DeepLearning.Layers
+{
+    public class AvgPooling2d : BHoMObject, IModule
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public Tuple<int, int> KernelSize { get; set; } = new Tuple<int, int>(2, 2);
```
adecler: How are Tuples working for you on the UI side? 
I don't remember us having any way to support them. If we don't, I wouldn't recommend a new component to support them as I woulid already like to get rid of the dictionary component. Can't we create a `Size2D` object instead ?

epignatelli: Yep, that sounds good!
Would something like the last commit (https://github.com/BHoM/BHoM/pull/586/commits/ac715186aa69e4b18c582f19205ff40075accb1c) work?

If so, I'll go through the rest of the objects.

adecler: Yes. Just one detail: Is there a better way to identify the dimensions than 1 and 2 ?

epignatelli: Yes, a shift is about to happen in the DL community about named axes: https://github.com/pytorch/pytorch/issues/4164

The drawback is that the logic may be leaky because the two dimensions can have different meanings. For example, I can have a 1-D vector that represents a word, and have a batch dimension (and you have `Batch` and `Width`); or I can have a single image with two dimensions, and then you have `Height` and `Width`. 

The way we could solve this is to used the `KeyedCollection`, the same way we used it for the `FragmentSet`. That way you can access the dimension by both name and position.
But I'd consider it an enhancement to the current pr, rather than solving it in this one..

epignatelli:
Merging this since this is experimental, and will go through a series of iterations before getting it right anyway.

epignatelli:
Thanks @al-fisher !

Issue #556, opened 2019-09-23
Structure_oM: Add Mode to results class


rwemay:
#### Description:

Add mode property (string) to the structural results class, to allow assignment of the mode the result has been extracted for, specifically/in particular for seismic response spectrum analysis cases. The property should be a string such that the method of combining the results can be stored, if not a specific mode (CQC, SRSS, 2SM etc.) for the cases where superposed results are requested/filtered.

PR #587, opened 2019-10-18
Structures o m #571 structural result requests

### LINKED ISSUE #571: BHoM_Adapter: refactoring-Structural_Adapter queries
Raising here to start the discussion about new IQueries to be implemented to be used by Structural_Adapters in general, and hopefully also other software adapters. Meant to be linked up with the adapter rewamp started by @adecler in https://github.com/BHoM/BHoM_Adapter/pull/70

The below is Structures focused, but good to get this as discipline agnostic as possible for parts where it might be possible, such as elements and misc queries. For results we might have to end up diverging, but good to keep it as open as possible, and at least try to agree a common strategy.

Some ideas so far:

### Results

Could group each query type with its corresponding element type. These could all contain settings that would match with different common ground extraction settings in the various packages we are using. For example (to be discussed and agreed):

- MeshResultQuery:
   - MeshResultType
   - MeshResultSmoothingType
   - MeshResultLayer
   - LayerPosition
   - CoordinateSystem
   - Cases
   - ObjectIds
- BarResultQuery:
   - BarResultType
   - Divisions
   - DivisionType (Evenly distributed vs ensure max values etc)
   - Axis (local/global) - unless we set a standard for this we _always_ use
   - Cases
   - ObjectIds
- NodeResultQuery:
   - NodeResultType
   - Axis (local/global) - unless we set a standard for this we _always_ use
   - Cases
   - ObjectIds
- PanelResultQuery:
   - PanelResultType
   - CoordinateSystem
   - CuttingType
   - Cases
   - ObjectIds
- GlobalResultQuery

### Elements

The filter query serves a good purpose in general, when you are trying to query out every single object of a specific type. Could be useful to have additional parameters such as being able to provide element ids for the things you are getting out.

Potential additional query:

- ElementQuery
   - Type
   - Ids
   - Tag

### Other types:

One query type @ZiolkowskiJakub has started to make use of in Revit_Toolkit, for now as a filter query, that could work some some other softwares as well is to be able to pull by selection. Also to get a system in place that could handle merging of multiple queries into one batch, and having something central to deal with this will be needed.

- SelectionQuery
- BatchQuery (existing, but needs more helper methods)

Any comments on the above welcome. Once agreed, will split into smaller chunk issues for implementation.

IsakNaslundBh:

Closes #571 

Adding Requests for Structural Results

### DIFF
--- a/Data_oM/Requests/IResultRequest.cs
+++ b/Data_oM/Requests/IResultRequest.cs
@@ -1,0 +1,36 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using BH.oM.Base;
+
+
+namespace BH.oM.Data.Requests
+{
+    public interface IResultRequest : IRequest
+    {
+        List<object> ObjectIds { get; set; }
+
+        List<object> Cases { get; set; }
+    }
+}
--- a/Structure_oM/Requests/BarResultRequest.cs
+++ b/Structure_oM/Requests/BarResultRequest.cs
@@ -1,0 +1,49 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+
+
+namespace BH.oM.Structure.Requests
+{
+    public class BarResultRequest : IStructuralResultRequest
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public BarResultType ResultType { get; set; } = BarResultType.BarForce;
+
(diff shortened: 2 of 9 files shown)


IsakNaslundBh:
Planning to add a Mode property to the structural result objects. https://github.com/BHoM/BHoM/issues/556

Hence added in the mode for the Requests straight away to avoid having to come back and change them later.

IsakNaslundBh:
`as per our chat, let's consider the options for the result Collection`

I see three options for this:

1. Make the `ReadResults(IResultRequest request)` method in the adapter return `IEnumerable<IObject>`. This way the return type can be either a `IResultCollection` or a `IResult`
1. Make `IResultCollection : IResult`. this way the ReadResults method could be returning an `IEnumerable<IResult>` as this could mean either single results or result collections.
1. Differentiate between `IResultRequest` and `IResultCollectionRequest` which fundamentally will be the same interface. Make the `MeshResultRequest : IResultCollectionRequest` instead. Then we have two methods in the BHoMAdapter. ReadResults and ReadResultCollection, returning `IEnumerable<IResult>` and `IEnumerable<IResultCollection>` respectively. 

I personally prefer option 2 of we can get that to work nicely, if not option 1. Happy to get other peoples comments on this as well.

al-fisher:
Of options above - do agree 1 or 2 would be preference. Do not want to impose differentiation of  `IResultRequest` and `IResultCollectionRequest` on user.

I think @IsakNaslundBh - option 2. is also my preferred. Considering any technical implications - let's discuss. _Now!!_ 😄

IsakNaslundBh:
After chat with @al-fisher and @alelom :

Option 2 from above seems like the best option and the way forward. This will require some tweaks to the IResultCollection as well as the MeshResult. Separate issue with thought for this opened here:
https://github.com/BHoM/BHoM/issues/589

See description in issue for details.

Data_oM/Requests/IResultRequest.cs
```diff
@@ -0,0 +1,36 @@
...
+using BH.oM.Base;
+
+
+namespace BH.oM.Data.Requests
+{
+    public interface IResultRequest : IRequest
+    {
+        List<object> ObjectIds { get; set; }
+
+        List<object> Cases { get; set; }
```
alelom: Not sure, but couldn't/shouldn't this be `List<ICase>` instead?

Structure_oM/Requests/BarResultRequest.cs
```diff
@@ -0,0 +1,49 @@
...
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public BarResultType ResultType { get; set; } = BarResultType.BarForce;
+
+        public DivisionType DivisionType { get; set; } = DivisionType.EvenlyDistributed;
+
+        public int Divisions { get; set; } = 5;
+
+        public List<object> Cases { get; set; } = new List<object>();
```
alelom: As comment above

Structure_oM/Requests/GlobalResultQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class GlobalResultQuery : IStructuralResultRequest
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public GlobalResultType ResultType { get; set; } = GlobalResultType.Reactions;
+
+        public List<object> Cases { get; set; } = new List<object>();
```
alelom: As comments above

Structure_oM/Requests/MeshResultRequest.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public MeshResultSmoothingType Smoothing { get; set; } = MeshResultSmoothingType.None;
+
+        public MeshResultLayer Layer { get; set; } = MeshResultLayer.AbsoluteMaximum;
+
+        public double LayerPosition { get; set; } = 0.0;
+
+        public Basis Orientation { get; set; } = Basis.XY;
+
+        public List<object> Cases { get; set; } = new List<object>();
```
alelom: As comments above

Structure_oM/Requests/NodeResultQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class NodeResultQuery : IStructuralResultRequest
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public NodeResultType ResultType { get; set; } = NodeResultType.NodeReaction;
+
+        public List<object> Cases { get; set; } = new List<object>();
```
alelom: As comments above

Data_oM/Requests/IResultRequest.cs
IsakNaslundBh: Reason for `List<object>` is to be able to feed in either `List<ICase>` as you stated OR for example `List<string>` or `List<int>` etc, so you either specify the id of the case, or use the object itself if that make sense?

alelom: Makes sense, thanks.

alelom:
@al-fisher this is ready to be merged!

PR #595, opened 2019-10-29
Data_oM: Add table class

### LINKED ISSUE #593: Data_oM: Add table class
#### Description:

Table class needed to be able to easily store and query tabulated data where the table itself is of interest and no suitable BHoMObejct exists.

First suggestion will be to use the DataTable class built into the framework.

IsakNaslundBh:

Closes #593 

Adding Table class

Using the C# DataTable to store the data. Adding additional MetaData proeprty to store things such as Design guidelines associated with the table.

### Additional comments

@Martian42


Data_oM/Collections/Table.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class Table : BHoMObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        public DataTable Data { get; set; }
+
+        public string MetaData { get; set; }
```
epignatelli: I am not sure I am up to date with this, so can't comment on the concept.
The only question is, what is `MataData`?

IsakNaslundBh: THanks @epignatelli , the Idea behind this was to be able to store complimentary information about the table. For example design assumptions made when generating the data in the table. string MetaData might not be the right name and type for it though, put that in for now as it felt flexible enough for now, but happy for other suggestions! @al-fisher @rwemay

epignatelli: Uhm, if it is volatile information, may it be useful to use `CustomData` directly?

rwemay:
This will be very handy for PostgreSQL!

Data_oM/Collections/Table.cs
IsakNaslundBh: Not sure if volatile information is correct way to describe it, but still think you might be right that this would be a better way to deal with this. To make it flexible enough I think the endpoint would be a `Dictionary<string,object>`, so then might as well use the CustomData.

IsakNaslundBh: Went for removing the property, will prototype with custom data when needed.

PR #596, opened 2019-10-30
Graphics_oM: Adding RenderMesh

### LINKED ISSUE #594: Geometry_oM: Update Mesh class to carry vertex colours
#### Description:

To enable visualisation of results on meshes, vertex colours needs to be added to the Mesh class.

Proposal to add new class called Vertex with Point and System.Drawing.Colour as properties and update the mesh class to carry a `List<Vertex>` instead of `List<Point>`.

Explicit casting from Point to Vertex should also be added for the Vertex class.

FYI @pawelbaran @enarhi @al-fisher @rwemay @FraserGreenroyd @adecler

kThorsager:
Closes #594 
This is all placed in the `Graphics_oM`

### Changelog

- Adds the Class `Vertex` which contains a `Point` and a `Color` 
- Adds the Class `RenderMesh` which contains a `List<Vertex>` and a `List<Face>` 

### Additional comments
Also set the default colour of the vertex to the same as in the Grasshopper_Toolkit


IsakNaslundBh:
Generally looks good to me. One quick question is if this should be implementing `IGeometry` or not? Think this makes sense, but want to give a comment on that as well @al-fisher ?

al-fisher:
> implementing `IGeometry` or not? 

Yep - agreed it should

al-fisher:
Just to note for the record @kThorsager 

This is an issue with CI - not your code 🎉 - is being fixed and I will retrigger. And merge when happy! 👍 

image

Issue #435, opened 2019-03-13
Reflection_oM: Add UrlAttribute for Assembly


epignatelli:
`[assembly: Url("https://github.com/BHoM/BHoM/")]` stores the url of the repo containing the assembly i.e. `Structure_oM` -> `[assembly: Url("https://github.com/BHoM/BHoM/")]`

PR #603, opened 2019-11-04
Geometry_oM: Extension of definition of NurbsSurface to handle trims

### LINKED ISSUE #425: Geometry_oM : Creation of Trimmed NurbsSurface definition
Definition of a trimmed Nurbs to be created in the oM (and the required converts to be raised in other toolkits) for interop purposes only.

As discussed and agreed https://github.com/BHoM/BHoM/issues/422#issuecomment-469591958 - focus certainly should be on representation in BHoM and not populating any Engine methods for manipulation. Host application can continue to be utilised for these purposes.

pawelbaran:

Closes #425

### Test files

Test files are located on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%2FGeometry%5FoM%2DIssue425%2DCreationOfTrimmedNurbsSurfaceDefinition). To run it, please switch to the branches related to the following PRs:
- [#603 ](https://github.com/BHoM/BHoM/pull/603)
- [#1319](https://github.com/BHoM/BHoM_Engine/pull/1319)
- [#122](https://github.com/BHoM/Rhinoceros_Toolkit/pull/122)
- [#435](https://github.com/BHoM/Grasshopper_Toolkit/pull/435)
- [#435](https://github.com/BHoM/Revit_Toolkit/pull/435)
- [#183](https://github.com/BHoM/Dynamo_Toolkit/pull/183)

The error in Rhino -> BHoM -> Rhino convert error in the test file is caused by the bug in Cone convert logged [here](https://github.com/BHoM/Rhinoceros_Toolkit/issues/121).

### Changelog

- `BH.oM.Geometry.SurfaceTrim` class has been added
- `BH.oM.Geometry.NurbsSurface` definition has been extended with additional properties in order to handle trims + `IImmutable` interface has been implemented

### Additional comments
This PR has been in progress for quite some time already - finally it got to the point where it makes sense to review it. In general, the definition of `NurbsSurface` has been extended in order to handle trims - I decided not to create a separate class for it because:
1. `PlanarSurface` contains the information about its outlines, therefore it seems logical to align `NurbsSurface` to it.
2. An unbounded nurbs surface will not often appear in AEC applications, and even if it will, the instance can always carry no trims.
3. Creation of another class would cause even more challenges on converts.

Both 3D curves and their representations in the UV coordinates of the surfaces are stored in `NurbsSurface`. The reason for that is the fact that even Rhino does not always handle conversion correctly ([Surface.Pullback](https://developer.rhino3d.com/api/RhinoCommon/html/M_Rhino_Geometry_Surface_Pullback.htm) and [Surface.Pushup](https://developer.rhino3d.com/api/RhinoCommon/html/M_Rhino_Geometry_Surface_Pushup.htm) methods often fail on seams). Division into outer and inner trims could possibly be avoided, but I believe it will make the code easier to apply. Pairs of 2d and 3d trim curves have been wrapped into `SurfaceTrim` class.

Additionally, the `NurbsSurface` class has been extended by two int properties: `UDegree` and `VDegree`. This is caused by the fact that querying the degree of a nurbs surface based on its knot vector is not always possible (see [#1297](https://github.com/BHoM/BHoM_Engine/issues/1297)).

### DIFF
--- a/Geometry_oM/Misc/SurfaceTrim.cs
+++ b/Geometry_oM/Misc/SurfaceTrim.cs
@@ -21,23 +21,30 @@
  */
 
 using BH.oM.Base;
-using BH.oM.Geometry;
-using BH.oM.Common.Properties;
-using BH.oM.Common.Interface;
+using System.Collections.Generic;
 
-using BH.oM.Reflection.Attributes;
-
-namespace BH.oM.Architecture.Elements
+namespace BH.oM.Geometry
 {
-    [Deprecated("2.3", "Replaced by BH.oM.Physical.Elements.Roof as part of migration to combined physical namespace")]
-    public class Roof : BHoMObject, IObject2D
+    public class SurfaceTrim : IGeometry, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        public Object2DProperties Properties { get; set; } = new Object2DProperties();
-        public ISurface Surface { get; set; } = new PlanarSurface();
+        public ICurve Curve3d { get; }
+
+        public ICurve Curve2d { get; }
+
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public SurfaceTrim(ICurve curve3d, ICurve curve2d)
+        {
+            Curve3d = curve3d;
+            Curve2d = curve2d;
+        }
 
         /***************************************************/
     }
--- a/Geometry_oM/Surface/NurbsSurface.cs
+++ b/Geometry_oM/Surface/NurbsSurface.cs
@@ -37,7 +37,19 @@ namespace BH.oM.Geometry
         public List<double> UKnots { get; set; } = new List<double>();
 
         public List<double> VKnots { get; set; } = new List<double>();
-        
+
+        public int UDegree;
+
+        public int VDegree;
+
+        public List<ICurve> ExternalBoundaries3d { get; set; } = new List<ICurve>();
+
+        public List<ICurve> ExternalBoundaries2d { get; set; } = new List<ICurve>();
+
+        public List<ICurve> InternalBoundaries3d { get; set; } = new List<ICurve>();
+
+        public List<ICurve> InternalBoundaries2d { get; set; } = new List<ICurve>();
+
         /***************************************************/
     }
 }
--- a/Geometry_oM/Surface/NurbsSurface.cs
+++ b/Geometry_oM/Surface/NurbsSurface.cs
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
@@ -22,6 +22,8 @@
 
 using BH.oM.Base;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
 
 namespace BH.oM.Geometry
 {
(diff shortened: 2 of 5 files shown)


al-fisher:
Thanks @pawelbaran  - hugely appreciate the thorough comments above. Really useful and addressed already a couple of the key questions I had on the changes you pushed in advance of raising PR! So actually with you comments I can agree with your approach. (Was imagining multiple classes - but augments make some sense).

In advance of testing and reviewing associated PRs etc. - one question/comment:

Might we now need to make `NurbsSurface` implement the `IImmutable` interface as there are now a couple of ways you can create self conflicting objects through combination of 2D and 3D boundaries and knot vectors and degrees?

pawelbaran:
> Might we now need to make `NurbsSurface` implement the `IImmutable` interface as there are now a couple of ways you can create self conflicting objects through combination of 2D and 3D boundaries and knot vectors and degrees?

Yes, I think so, particularly taken into account the number of mathematical requirements that need to be met to create a valid trimmed surface.

pawelbaran:
Good thoughts @epignatelli, 

> * Can we merge `*Boundaries2d` and `*Boundaries3d`? it is ok to keep the internal and external separate, even if it would be ideal to merge them.

Possibly we could be able to merge these two, but:
- if we leave only `*Boundaries3d`, it is problematic to degrade them into 2d curves on seams (at least I could not make Rhino handle that correctly)
- if we leave only `*Boundaries2d`, we would need to create our own 2d -> 3d converts, which I am not sure know how to approach (happy to implement though, if pointed to the right source)

Merging the inner and outer trims could be possible, but then we need a method to reverse that process - one property less, one method more. I am fine with both.

> * I followed the problem with getting the right degree from [BHoM/BHoM_Engine#1297](https://github.com/BHoM/BHoM_Engine/issues/1297). I think @enarhi did flag this a while ago, he may know something more about it. The question is, can we solve _that_ problem instead of storing degrees into the object itself? Why can't we have full multiplicity in the knot vector all the time?

I have replied to your comment in [#1297](https://github.com/BHoM/BHoM_Engine/issues/1297) - naturally, my first idea was to refactor the existing method instead of adding an extra property - I failed 😭

> * If the object is becoming this complex I agree with @al-fisher that it must implement the `IImmutable` interface. At this point I'd suggest to make it a brand new object `TrimmedSurface` and leave the `NurbsSurface` as it is. Otherwise a lot of logic we currently have in the engine needs to be revised.

I extended the existing class instead of adding a new one only because it was faster to start with. Then decided to follow that scheme when I realized that having two classes would add another layer of complexity to the converts. Finally, `PlanarSurface` also carries the information about its trims, therefore adding these to `NurbsSurface` sounds logical to me.

Could you point to the logic that would need to be revised? Happy to look into.

pawelbaran:
Following today's catchup with @al-fisher and @epignatelli: we will continue with the currently proposed definition of a `NurbsSurface`, actions to be taken before merging:
- [ ] add implementation of `IImmutable`
- [ ] get rid of hacks and shortcuts
- [ ] finish Revit converts
- [ ] reopen the PRs with same branch names

TBC.

rwemay:
I like point 2!

al-fisher:
Thanks @pawelbaran 
Discussion with @epignatelli was really useful. Cheers both. 
Summary for others - priority is being placed on robust and reliable converts and representation of the full breadth of NURBS Surfaces in BHoM - untrimmed/trimmed, periodic/non-periodic in both U or V etc.

Modify Engine methods or not priority for BHoM - with external libraries being hugely more powerful here naturally - hence focus on Engine Converts to and from BHoM.

To close out this suite of PRs - completion of converts to and from Rhinoceros and Revit are being used.

FraserGreenroyd:
As an FYI: I updated the branch for @pawelbaran because it did not contain the `MEP_oM` while his Engine PR did contain `MEP_Engine`, and was resulting in failed checks on the engine. The changes I made to the branch here should resolve the checks on the engine. Please note you may need to delete the branch from your local repo and refetch this branch if you had it checked out already! :smile:

Geometry_oM/Surface/NurbsSurface.cs
```diff
@@ -37,7 +38,19 @@ public class NurbsSurface : ISurface
         public List<double> UKnots { get; set; } = new List<double>();
 
         public List<double> VKnots { get; set; } = new List<double>();
-        
+
+        public int UDegree;
```
FraserGreenroyd: This should have `get` and `set` methods should it not?

Geometry_oM/Surface/NurbsSurface.cs
```diff
@@ -37,7 +38,19 @@ public class NurbsSurface : ISurface
...
+
+        public int VDegree;
```
FraserGreenroyd: This should have `get` and `set` methods should it not?

Geometry_oM/Surface/NurbsSurface.cs
IsakNaslundBh: Should have `get` methods only if we want to make this actually immutable. WIll put additional comments

pawelbaran:
Aye, good points @IsakNaslundBh! This will also induce changes in dependent repos - I will try to refactor the code today.

And what about the coupled 2d and 3d curves - would you guys keep them as separate lists or rather create a wrapper `SurfaceTrim` class or similar?

Geometry_oM/Surface/NurbsSurface.cs
epignatelli: I feel like my review would be redundant, so commenting here just to tick the box.

pawelbaran: Fixed

Geometry_oM/Surface/NurbsSurface.cs
pawelbaran: Fixed

pawelbaran:
The PR is ready to merge from my point of view.

Geometry_oM/Misc/SurfaceTrim.cs
```diff
@@ -0,0 +1,51 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using System.Collections.Generic;
+
+namespace BH.oM.Geometry
+{
+    public class SurfaceTrim : IGeometry, IImmutable
+    {
```
al-fisher: Like this concept now of a `SurfaceTrim` object combining the issues of 2D and 3D representation as a single `Immutable` object.

al-fisher:
This is looking good @pawelbaran - thanks - will follow the chain through with some further testing as discussed! 👍

PR #604, opened 2019-11-05
Update the README installer link

al-fisher:
Legacy installer hyperlink is updated to http://bhom.xyz

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A great place to start is reading our Wiki [here](https://github.com/BHoM/docume
 
 ## Quick start ##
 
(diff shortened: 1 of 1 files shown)


epignatelli:
Hi @al-fisher , @FraserGreenroyd.
Not sure if it was deliberate to point to bhm.xyz rather than to the installer itself.

The installer now (https://github.com/BHoM/bhom.github.io/pull/22) has a unique url that we can point to. The url is `https://bhom.xyz/installer/`.

Everytime we update the installer it automatically redirects to the latest one.

PR #659, opened 2020-01-17
BHoM: Added null check to `FragmentSet` constructor with `fragmentSet` argument.

### LINKED ISSUE #656: BHoM: FragmentSet copy constructor needs a nullcheck
#### Description:

Found by @kThorsager here https://github.com/BHoM/BHoM_Engine/pull/1420#pullrequestreview-344656646 the copy constructor fails for empty fragmentsets as the inner dictionary is null for an anpty `FragmentSet`. The copy constructor hence need a null check for this.

#### How to replicate:

Try calling the copy constructor with an empty fragment set

#### Expected behaviour:

For the copy constructor to not crash and generate a valid FragmentSet

alelom:

Closes #656 

As title.

### DIFF
--- a/BHoM/FragmentSet.cs
+++ b/BHoM/FragmentSet.cs
@@ -35,7 +35,8 @@ namespace BH.oM.Base
 
         public FragmentSet(FragmentSet fragmentSet)
         {
-            fragmentSet.Dictionary.Values.ToList().ForEach(v => this.Add(v));
+            if (fragmentSet != null && fragmentSet.Dictionary != null && fragmentSet.Dictionary.Count != 0)
+                fragmentSet.Dictionary.Values.ToList().ForEach(v => this.Add(v));
         }
 
         public bool AddOrReplace(IBHoMFragment fragment) // Slower than Add() or SetItem(), but easier to use.


al-fisher:
@alelom @kThorsager 

I think we are still getting a similar or related problem through the default `FragmentSet` Constructor - `FragmentSet()` - which is getting called as part of the object initializer.

I am getting a full on Rhino crash.

You can recreate by trying to create a `FragmentSet` through the create component - choosing default initializer option.

image

al-fisher:
I have not requested changes as I think is a related but separate bug ? - but probably worth you both having a quick look to see if makes sense to include in this PR fix - thanks

alelom:
Thanks @al-fisher !
This actually prompts a question: do we want to expose that constructor for `FragmentSet` (which is automatically generated)? 
I would say no, because it basically is exposing _the most complete base constructor_ of the class `KeyedCollection` of ` System.Collections.ObjectModel`, which includes these parameters:
- `IEqualityComparer<TKey> Comparer`
- `int dictionaryCreationThreshold`

while the third parameter visible in that automatically generated constructor is from our `BH.oM.Base`:
- IBHoMFragment fragment

Now honestly I didn't expect this to happen – I didn't fully understand the range/mechanism of our constructor generator. But I believe this is not a constructor we want to expose to users. Possibly, the mixing of these parameters is actually even causing the crash?

Do we have a way to prevent the generation of such constructors?

alelom:
I raised https://github.com/BHoM/BHoM_Engine/issues/1568 to cover the comments above.
I think this is good to be merged.

al-fisher:
Thanks @alelom

PR #662, opened 2020-01-21
NotExposedAttribute

### LINKED ISSUE #661: Reflection_oM: NotExposedAttribute
#### Description:

Would like to not expose things in the UI without making them private/internal

kThorsager:

Closes #661 
A attribute which hides a method from the UI

### Test files

Test any file to make sure it doesn't break anything.
But as this PR is not for setting anything as `NotExposed` no proper test file will be provided for that, instead set a current method as `NotExposed` and test that it disappeared from the UI

### DIFF
--- a/Reflection_oM/Attributes/NotExposedAttribute.cs
+++ b/Reflection_oM/Attributes/NotExposedAttribute.cs
@@ -1,0 +1,54 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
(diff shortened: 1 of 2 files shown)


alelom:
@al-fisher would you mind having a look and a test go too?

alelom:
I get your point @IsakNaslundBh -- thought about that after approving actually.

al-fisher:
> Code and UI being one to one is one of the core pillars of what we are doing, and this attribute is going straight against that. 

Agreed with above and @IsakNaslundBh your other comments 
Had a good chat just now with @alelom @kThorsager. 

One thing to capture is we do need to enable granularity of control of which bit of code is loaded/reflected into UIs or certain menus - this  functionally however should be : 
1. user controlled - i.e. what I want exposed may well be different to what you want exposed. 
Such that should be a user/ui config rather than hardcoded attribute. 

and 

2. perhaps better served at toolkit/(sub)namespace level and thus different mechanism to attributes for reasons above.

PR #664, opened 2020-01-22
Data_oM: Add attribute dictionary to GraphNode and GraphLink

### LINKED ISSUE #633: Data_oM : Generalise the attribute on a GraphLink beyond simple Weight parameter
Will be useful to generalise the approach to assigning attributes to `BH.oM.Data.Collections.Graph` entities (both `Nodes` as well as `Links` actually) beyond simple weightings.

rolyhudson:
Adding a generic dictionary for both node and edge attributes. Much like those used in NetworkX https://networkx.github.io/documentation/stable/reference/glossary.html#term-node-attribute
https://networkx.github.io/documentation/stable/reference/glossary.html#term-edge-attribute 

Closes #633 
Just edging this forward and hoping to get back to the discussion. 

### Additional comments
Also relates to:
https://github.com/BuroHappoldEngineering/Qualia_Toolkit/pull/5
https://github.com/BuroHappoldEngineering/Qualia_Toolkit/issues/3
Plus:
Already I'm using decision graphs for the venue performance rating work. This proposed change or similar will help massively. 
A weighting decision graph works using the existing BHoM Graph but its also nice to store scores as a graph. This allows querying the graph for the score at any sub node without needing to compute the score. 
I would also like to bring to the discussion the need for graph methods. I have implemented a method for depth first graph traversal and building an adjacency list both out of necessity for the Scoring using a decision graph. These are generic methods so should be in another namespace ultimately.
I know these additional comments are :rewind: sprawling ⏩  :octopus: but think I need to get it all down somewhere and get some critical 👀 eyes on it 😃 .

### DIFF
--- a/Data_oM/Collections/GraphLink.cs
+++ b/Data_oM/Collections/GraphLink.cs
@@ -19,6 +19,7 @@
  * You should have received a copy of the GNU Lesser General Public License     
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
+using System.Collections.Generic;
 
 namespace BH.oM.Data.Collections
 {
@@ -28,7 +29,7 @@ namespace BH.oM.Data.Collections
         /**** Properties                                ****/
         /***************************************************/
 
-        public double Weight { get; set; } = 1.0;
+        public Dictionary<string,object> Attributes { get; set; } = new Dictionary<string, object>();
 
         public GraphNode<T> StartNode { get; set; } = new GraphNode<T>();
 
--- a/Data_oM/Collections/GraphLink.cs
+++ b/Data_oM/Collections/GraphLink.cs
@@ -30,11 +30,9 @@ namespace BH.oM.Data.Collections
         /***************************************************/
 
         public Dictionary<string,object> Attributes { get; set; } = new Dictionary<string, object>();
-
+        public double Weight { get; set; } = 1.0;
         public GraphNode<T> StartNode { get; set; } = new GraphNode<T>();
-
         public GraphNode<T> EndNode { get; set; } = new GraphNode<T>();
-
 
         /***************************************************/
     }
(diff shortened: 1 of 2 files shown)


epignatelli:
Removing from reviewer as it is still in WIP and appears in my to-do list  but can't do anything to solve it :D

@rolyhudson would you mind tagging me whenever this is ready for review?

adecler:
Adding an `Attributes` dictionary makes sense but I would keep Weight as a separate property. This would guarantee a weight property is always available and would better mirror the fact that Value is kept separated in Node.

rolyhudson:
@adecler cheers for the suggestion sounds good although I don't understand why weight is of particular importance?
@epignatelli - sorry for the ignorant ``WIP`` tag use again would appreciate you review and thoughts

adecler:
@rolyhudson , many graph algorithms rely on the node having a value and the links having a weight/value. Having the weight stored in a dictionary means that any code using it would have to constantly check that that value is even contained in the dictionary and probably fail if it isn't.

rolyhudson:
Thanks for the comments @epignatelli and @adecler.
Will check the engine @epignatelli is this why we get ``BHoM.CheckCore Failing after 1m — Build #20200122.1 failed``?
Also thanks for the explanation notes on graph algorithms @adecler makes good sense.

epignatelli:
> Will check the engine @epignatelli is this why we get `BHoM.CheckCore Failing after 1m — Build #20200122.1 failed`?

Yep! That's it!

rolyhudson:
Also see the pull request https://github.com/BHoM/BHoM_Engine/pull/1449

rolyhudson:
> Hey @rolyhudson no worries at all!
> 
> I believe we need a corresponding pr in the engine though, as it breaks some code.
> Check this out to make sure the CI coordinates the prs properly: https://github.com/BHoM/documentation/wiki/Resolving-an-Issue#in-particular-note-branches-in-dependant-repos---must-be-named-identically

I think this is now handled correctly @epignatelli ?

al-fisher:
> I think this is now handled correctly @epignatelli ?

Perfect! Can see BHoM.CheckCore CI check is happy now! Thanks @rolyhudson 

Issue on other PR - but think that this is actually a CI problem! Sorry
See comment here: https://github.com/BHoM/BHoM_Engine/pull/1449#issuecomment-581580950

PR #679, opened 2020-01-31
Environment_oM : Add IBHoMFragment to EnvironmentMaterial Fragments

### LINKED ISSUE #675: Environment_oM: Environment materials need to implement IBHoMFragment
#### Broken rules:

Similar to https://github.com/BHoM/BHoM/issues/588, `IEnvironmentMaterial` need to implement `IBHoMFragment`. As soon as done, the other interface can be removed and physical material updated to rely on fragments instead of `IMaterialProperties`

FraserGreenroyd:
Fixes #675 

### Test files
N/A

### DIFF
--- a/Environment_oM/MaterialFragments/GasMaterial.cs
+++ b/Environment_oM/MaterialFragments/GasMaterial.cs
@@ -32,7 +32,7 @@ using System.ComponentModel;
 namespace BH.oM.Environment.MaterialFragments
 {
     [Description("Fragment containing the Gas Material properties related to materials")]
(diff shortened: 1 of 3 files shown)


al-fisher:
> some extent only needed on the base interface, but think it can be sensible as a explicit addition on the actual materials as well.

Here I would agree - as `IEnvironmentMaterial` does not have “Fragment” in its name so helpful on `GasMaterial` to be explicit to fact is a fragment!

Might be nice to formalise guidance 
@adecler

PR #684, opened 2020-02-04
Reflection_oM: Adding quantity attributes to inputs and outputs

### LINKED ISSUE #683: Reflection_oM: Add quatity attributes to Input and Output attributes
#### Description:

Add QuantityAttribute as a property of the input and output attributes to be able to decorate them with this as well. Constructor will need to accept type due to restrictions.

IsakNaslundBh:

Closes #683 

Adding Quantity attributes to input and output attributes. This will let us decorate input parameters and output parameters by the use of `typeof(Length)` for example.

Reason for accepting type rather than a Quantity attribute in the constructor is due to limitations on what types that can be passed to a Attribute constructor. @adecler Good to check with you if you see any problems with this?
This will mean that we will lose typestrongness but the syntax will actually look better, `typeof(Length)` over `new Length()`.

### DIFF
--- a/Reflection_oM/Attributes/InputAttribute.cs
+++ b/Reflection_oM/Attributes/InputAttribute.cs
@@ -26,6 +26,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Reflection.Attributes
 {
@@ -40,6 +41,7 @@ namespace BH.oM.Reflection.Attributes
 
         public string Description { get; private set; } = "";
 
+        public QuantityAttribute Quantity { get; set; } = null;
 
         /***************************************************/
         /**** Constructors                              ****/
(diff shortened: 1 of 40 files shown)


Reflection_oM/Attributes/InputAttribute.cs
```diff
@@ -51,6 +53,17 @@ public InputAttribute(string name, string description)
             Description = description;
         }
 
+        /***************************************************/
+
+        public InputAttribute(string name, string description, Type quantity)
```
FraserGreenroyd: Do we need a separate constructor for this? Can't we just modify the existing one to have `quantity = null` as a default given you're doing the `if` check in this one? I'm not overly concerned so not requesting changes - just my 'off the top of my head' thought while combing through

IsakNaslundBh: Think you are right here. Did it this way to be as non-intrusive as possible to what we already had, but might make more sense to just add it in with `quantity= null` as default.

Any other thoughts @al-fisher @adecler ?

al-fisher: Agreed

al-fisher:
For what it’s worth, I really like the `Typeof(Length)` syntax. Post-rationalising 😄

rwemay:
:)

Reflection_oM/Attributes/InputAttribute.cs
adecler: me too

adecler:
You guys have done all the work and I like the result. I'm happy 😄

PR #687, opened 2020-02-05
Environment_oM: AnalyticalConstruction fragment

### LINKED ISSUE #615: Environment_oM: Calculated construction values fragment
#### Description:

It would be useful to have the heavily calculated values from energy modelling software read and stored on constructions as fragments. These include: U-value, g-value, thermal mass Cm, visible light transmission etc.

MajaLindroth:

Closes #615 

AnalyticalConstruction gets calculated values from modelling software and store them as fragments.


Environment_oM/Fragments/AnalyticalConstruction.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Analytical.Elements;
+
+using System.ComponentModel;
+
+namespace BH.oM.Environment.Fragments
+{
+    public class AnalyticalConstruction : IBHoMFragment, IAnalyticalConstruction, IImmutable
```
IsakNaslundBh: This class is not immutable ( not sure if that is the intention or not)

If this is meant to be the version of the panel property where you can set all of those properties I would keep the class as it is but maybe rename containing the word explicit. If this is done, then should also remove the IImmutable interface.

If you aim to go down the same route as we have done for the SectionProperties, I would also add a base interface with all of these properties as get only.

IsakNaslundBh: Happy to have a chat about how we have done it in structures if that would help.

Issue #645, opened 2020-01-10
Data_oM: BatchRequest vs LogicalAndRequest and LogicalOrRequest


pawelbaran:
While reviewing the requests in Revit_Toolkit, I got stuck at two:
- `LogicalAndRequest`
- `LogicalOrRequest`

They seem to be generic enough to sit in a more central repo than Revit_Toolkit. A few questions come to my mind:
- Would there be a use case for that? For example pull only bars that are made of certain profile?
- Currently both requests are handled internally by Revit_Toolkit - if they were moved to BHoM, how much work would you expect to be needed in each/your adapter?
- How does `BatchRequest` correspond to the ones above?

IsakNaslundBh:
Moving them centrally would get an upvote from me. For sure concepts that are generic anough to live outside revit toolkit!

> Would there be a use case for that? For example pull only bars that are made of certain profile?

Yes!

> Currently both requests are handled internally by Revit_Toolkit - if they were moved to BHoM, how much work would you expect to be needed in each/your adapter?

Would be fine with them not being supported straight away out of the box. `BatchRequest` is not supported in any of the structural adapters as it stands for example. Kan then get implemented on a toolkit by toolkit basis as needed.

And not sure about the batch request

al-fisher:
Agreed with all the above! 👍 

Can return generic error stating request type not supported by this adapter?

alelom:
Agree with the move as well. Can definitely see them used outside revit.

adecler:
Agreed too. Mongo definitely comes to mind

pawelbaran:
Cool, I will move the two to Data.oM together with their relevant methods. However, could anyone tell me whether they overlap with `BatchRequest` by any means? Do we have any use case of `BatchRequest` anywhere? Thanks!

adecler:
That I know of:

- https://github.com/BHoM/Mongo_Toolkit/blob/master/Mongo_Adapter/AdapterActions/Pull.cs
- https://github.com/BHoM/Mongo_Toolkit/blob/master/Mongo_Adapter/AdapterActions/Push.cs

pawelbaran:
This issue can be closed - the logical requests got added.

PR #688, opened 2020-02-06
Data_oM: LogicalAnd and LogicalOr requests added

pawelbaran:

This PR is resolving #645 to some extent, but happy to keep the issue open as it would be good to make sure that `LogicalOrRequest` does not do exactly same thing as `BatchRequest`.

### DIFF
--- a/Data_oM/Data_oM.csproj
+++ b/Data_oM/Data_oM.csproj
@@ -57,10 +57,13 @@
     <Compile Include="Requests\BatchRequest.cs" />
     <Compile Include="Requests\CustomRequest.cs" />
     <Compile Include="Requests\FilterRequest.cs" />
+    <Compile Include="Requests\ILogicalRequest.cs" />
     <Compile Include="Requests\IRequest.cs" />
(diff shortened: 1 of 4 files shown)


pawelbaran:
@FraserGreenroyd `ILogicalRequests` will enable building logical queries. Let me explain this on an example: the user may want to pull e.g. only Revit elements of certain type that belong to a certain workset. Such user could:
a) pull all element of given type and then filter them by workset
b) use `LogicalAndRequest`

The advantage of the latter is that it does not require converting all objects, but only the ones of interest, additionally it makes things much easier for the end user. Looking into the original issue may also spot some light on the idea.

adecler:
I am fine with this. 
Probably good to make sure you coordinate with @rwemay  as he is looking into very similar things at the moment.
Also worth pinging @emidio-piermarini for his early work on Mongo queries.

pawelbaran:
@al-fisher @rwemay could we get the final blessing on this one from one of you and merge? Thanks!

Issue #636, opened 2019-12-13
Adding a DataPoint object to Data_oM


rolyhudson:
Following discussions with @IsakNaslundBh, @rwemay and @jdarcyBHE can we consider adding a `DataPoint` object?
This would be a more granular version of `Dataset` in order to allow a single value to attributed to a source. Something like:
```
public class DataPoint : BHoMObject
{
    /***************************************************/
    /****            Public Properties              ****/
    /***************************************************/
    public object Value { get; set; } = new object();
    public Source SourceInformation { get; set; } = new Source();
    public DateTime TimeOfCreation { get; set; }
    /***************************************************/
}
```
Working on the venue performance rating system the Dataset associated with each venue comes from multiple sources sometimes where Capacity for one venue comes from sourceA and from sourceB for other venues.
What are your thoughts?

rolyhudson:
Do we also need an interface `IData` that would ensure the properties of 
```
public Source SourceInformation { get; set; }
public DateTime TimeOfCreation { get; set; }
```
Are included in DataPoint and Dataset?

al-fisher:
Initial thoughts - and talking with @rolyhudson just now too - I am wondering if we keep simple with our current single object `Dataset` but clarify and encourage that Property `List<IBHoMObject> Data` can be a _set of one_ if only a single data _point_.
That said - I am not against creating another object for single items - if querying say `.Value` makes more sense for a user and really helps the humanising of the concepts!
Suggest we do next iteration on class diagrams and including the new ideas of `IData`/`IMetric` interfaces - to help make decision also.
@rwemay @IsakNaslundBh @rolyhudson

rwemay:
Summary of discussion with @al-fisher 

 - Sounds like having a `Data` class rather than `DataPoint' could be worthwhile, and `Data` would have `Source` and `Value` and has to implement `IComparable` such that the value can be compared (but does not need to be primitive, as long as what ever is being compared has `IComparable`. 

 - The dictionary inside the fragment attached to any object would then be `Dictionary<string,data> Data`. 

 - Engine methods would be set up to `Extract` the `Data` from any object, as well as any necessary reconciliation of Data versus formal object properties. 

 - Engine methods would be used to convert `Data` into a `Metric` and/or a `Metricset`. `Metric` will own `Value` - the value of the data, `Range` - the range of possible values the data can take (can be set at construction or updated based on a dataset, e.g. range of pitch lengths according to EPL.

[...]

PR #693, opened 2020-02-07
Data_oM #636 adding data point object

rolyhudson:
Just to return to the conversation: https://github.com/BHoM/BHoM/issues/636#
DataPoint has become a key object for Performance Rating and been tested on the Venue Performance Rating League Tables project.


Data_oM/Library/DataPoint.cs
```diff
@@ -0,0 +1,51 @@
+﻿using BH.oM.Base;
+using BH.oM.Data.Library;
+using System;
+using System.Collections.Generic;
+
+namespace BH.oM.Data.Library
+{
+    public class DataPoint : BHoMObject, IComparable<DataPoint>
```
epignatelli: Have you already discarded the option of using a generic type `DataPoint<T>`

This should void the `IComparable` interface and delegate it to the specific type, allowing anyone to implement a speficic comparison it their objects, rather than here.

epignatelli: Rather than a `CompareTo`, I would add instead an `implicit` casting to the `T` type, such that you mihgt use `DataPoint<int>` as an `int` straight away.

rolyhudson: Yes you are correct. Thanks @epignatelli  I have completely overlooked the benefit of using generics here. Switching this to ``WIP`` while I revisit.

epignatelli: Oh no, I was just asking actually - not sure what the plan was!

rolyhudson: @epignatelli I think this could remove some painful issues. Great point 👍

pawelbaran:
@rolyhudson could you ping me once this one is merged? I would like to push #702 when possible.

rolyhudson:
@pawelbaran sorry I didn't realise this was holding you up. Why is that?
I need to revisit the definition of this DataPoint specifically if it is better to define as DataPoint<T>.
Do you have an opinion on that?

pawelbaran:
@rolyhudson I have no opinion on the subject you are dealing with in this PR - I do not know what is it about tbh :-) The conflict comes from the fact that both your and my PR will attempt to change the same file (_Data_oM.csproj_).

If you expect your work to take a time counted in weeks rather than days, i would be happy to cut into the line and push mine in the meantime - this will require changing one line in you code to eliminate the conflict. Otherwise I can wait. What would you prefer?

rolyhudson:
Thanks @pawelbaran think I'm going to close this while I reconsider.

PR #701, opened 2020-02-10
Geometry_oM: Adding description and quantity attributes 

### LINKED ISSUE #700: Geometry_oM: Addition of description attributes to Geometry_oM
Addition of Description Attribute is required across Geometry_oM

al-fisher:

Part of #700 not complete yet. So not closing

### Additional comments

~Still in `WIP` - working my way through~
Taking of `WIP` for review and merge now

### DIFF
--- a/Geometry_oM/CoordinateSystem/Cartesian.cs
+++ b/Geometry_oM/CoordinateSystem/Cartesian.cs
@@ -25,19 +25,23 @@ using BH.oM.Base;
 
 namespace BH.oM.Geometry.CoordinateSystem
 {
-    [Description("Cartesian coordinate system. All vectors to be orthogonal unit vectors")]
+    [Description("A three-dimensional Cartesian coordinate system, as defined by a set of mutually orthogonal unit vectors and a common origin point")]
     public class Cartesian : IGeometry, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Unit vector defining X coordinate axis")]
         public Vector X { get; } = Vector.XAxis;
 
+        [Description("Unit vector defining Y coordinate axis")]
         public Vector Y { get; } = Vector.YAxis;
 
+        [Description("Unit vector defining Z coordinate axis")]
         public Vector Z { get; } = Vector.ZAxis;
 
+        [Description("Location point in three-dimensional space defining (0,0,0) and the origin for each axis")]
         public Point Origin { get; set; } = Point.Origin;
 
         /***************************************************/
--- a/Geometry_oM/CoordinateSystem/Cartesian.cs
+++ b/Geometry_oM/CoordinateSystem/Cartesian.cs
@@ -32,13 +32,13 @@ namespace BH.oM.Geometry.CoordinateSystem
         /**** Properties                                ****/
         /***************************************************/
 
-        [Description("Unit vector defining X coordinate axis")]
+        [Description("Unit vector defining local x coordinate axis")]
         public Vector X { get; } = Vector.XAxis;
 
-        [Description("Unit vector defining Y coordinate axis")]
+        [Description("Unit vector defining local y coordinate axis")]
         public Vector Y { get; } = Vector.YAxis;
 
-        [Description("Unit vector defining Z coordinate axis")]
+        [Description("Unit vector defining local z coordinate axis")]
         public Vector Z { get; } = Vector.ZAxis;
 
         [Description("Location point in three-dimensional space defining (0,0,0) and the origin for each axis")]
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -20,26 +20,31 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
-using System.ComponentModel;
 using System;
-
+using System.ComponentModel;
+using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("Arc starting from the local x, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. Arc starting from the local x, rotating Angle number of radians counter clockwise")]
     public class Arc : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Defines the Arc centre as the CoordinateSystem Origin, as well as the orientation of the Arc in three-dimensional space, lying in the CoordinateSystem XY-Plane")]
         public CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new CoordinateSystem.Cartesian();
 
+        [Length]
+        [Description("Distance from the CoordinateSystem.Origin to the Arc boundary")]
         public double Radius { get; set; } = 0;
 
+        [Angle]
         [Description("Angle in radians to the start point from the local X-axis, counter clowise around the local Z-axis")]
         public double StartAngle { get; set; } = 0;
 
+        [Angle]
         [Description("Angle in radians to the end point from the local X-axis, counter clowise around the local Z-axis")]
         public double EndAngle { get; set; } = 0;
 
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -26,26 +26,26 @@ using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. A circular Arc, a curve, with a constant distance for a point, its centre, defined it the local angular domain between start and end angle.")]
+    [Description("A plane curve. A circular Arc, a curve, with a constant distance from a point, its centre, defined it the local angular domain between start and end angle.")]
     public class Arc : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        [Description("Defines the Arc centre as the CoordinateSystem Origin, as well as the orientation of the Arc in three-dimensional space, lying in the CoordinateSystem XY-Plane")]
+        [Description("Defines Arc centre equal to Origin of the CoordinateSystem, as well as the orientation of the Arc in three-dimensional space, equal to local xy-Plane of the CoordinateSystem.")]
         public CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new CoordinateSystem.Cartesian();
 
         [Length]
-        [Description("Distance from the CoordinateSystem.Origin to the Arc boundary")]
+        [Description("Distance from the CoordinateSystem.Origin to the any point on Arc.")]
         public double Radius { get; set; } = 0;
 
         [Angle]
-        [Description("Angle in radians to the start point from the local x-axis, counter clowise around the local z-axis")]
+        [Description("Angle in radians to the start point from the local x-axis, counter clockwise around the local z-axis.")]
         public double StartAngle { get; set; } = 0;
 
         [Angle]
-        [Description("Angle in radians to the end point from the local x-axis, counter clowise around the local z-axis")]
+        [Description("Angle in radians to the end point from the local x-axis, counter clockwise around the local z-axis.")]
         public double EndAngle { get; set; } = 0;
 
         /***************************************************/
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -26,7 +26,7 @@ using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. Arc starting from the local x, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. Arc starting from the local X, rotating Angle number of radians counter clockwise")]
     public class Arc : ICurve
     {
         /***************************************************/
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -26,7 +26,7 @@ using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. Arc starting from the local X, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. A circular Arc, a curve, with a constant distance for a point, its centre, defined it the local angular domain between start and end angle.")]
     public class Arc : ICurve
     {
         /***************************************************/
@@ -41,11 +41,11 @@ namespace BH.oM.Geometry
         public double Radius { get; set; } = 0;
 
         [Angle]
-        [Description("Angle in radians to the start point from the local X-axis, counter clowise around the local Z-axis")]
+        [Description("Angle in radians to the start point from the local x-axis, counter clowise around the local z-axis")]
         public double StartAngle { get; set; } = 0;
 
         [Angle]
-        [Description("Angle in radians to the end point from the local X-axis, counter clowise around the local Z-axis")]
+        [Description("Angle in radians to the end point from the local x-axis, counter clowise around the local z-axis")]
         public double EndAngle { get; set; } = 0;
 
         /***************************************************/
--- a/Geometry_oM/Curve/Circle.cs
+++ b/Geometry_oM/Curve/Circle.cs
@@ -32,7 +32,7 @@ namespace BH.oM.Geometry
         /**** Properties                                ****/
         /***************************************************/
 
-        [Description("Origin point defining location of the circle in three-dimensional space")]
+        [Description("Origin point defining location of the Circle in three-dimensional space")]
         public Point Centre { get; set; } = new Point();
 
         [Description("Vector perpendicular to the plane in which the Circle lies in. Defines the orientation in three-dimensional space")]
--- a/Geometry_oM/Curve/Circle.cs
+++ b/Geometry_oM/Curve/Circle.cs
@@ -25,21 +25,21 @@ using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. Standard circle defining a curve of constant distance from a point, its Centre")]
+    [Description("A plane curve. Standard circle defining a curve of constant distance from a point, its Centre.")]
     public class Circle : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        [Description("Origin point defining location of the Circle in three-dimensional space")]
+        [Description("Origin point defining location of the Circle in three-dimensional space.")]
         public Point Centre { get; set; } = new Point();
 
-        [Description("Vector perpendicular to the plane in which the Circle lies in. Defines the orientation in three-dimensional space")]
+        [Description("Vector perpendicular to the plane in which the Circle lies.")]
         public Vector Normal { get; set; } = new Vector { X = 0, Y = 0, Z = 1 };
 
         [Length]
-        [Description("Distance from the Centre to the Circle boundary")]
+        [Description("Distance from the Centre to any point on Circle.")]
         public double Radius { get; set; } = 0;
 
         /***************************************************/
--- a/Geometry_oM/Curve/Circle.cs
+++ b/Geometry_oM/Curve/Circle.cs
@@ -20,18 +20,26 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
+using BH.oM.Quantities.Attributes;
+
 namespace BH.oM.Geometry
 {
+    [Description("A plane curve. Standard circle defining a curve of constant distance from a point, its Centre")]
     public class Circle : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Origin point defining location of the circle in three-dimensional space")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Vector perpendicular to the plane in which the Circle lies in. Defines the orientation in three-dimensional space")]
         public Vector Normal { get; set; } = new Vector { X = 0, Y = 0, Z = 1 };
 
+        [Length]
+        [Description("Distance from the Centre to the Circle boundary")]
         public double Radius { get; set; } = 0;
 
         /***************************************************/
--- a/Geometry_oM/Curve/Ellipse.cs
+++ b/Geometry_oM/Curve/Ellipse.cs
@@ -20,22 +20,33 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
+ using BH.oM.Quantities.Attributes;
+
 namespace BH.oM.Geometry
 {
+    [Description("A plane curve. A standard ellipse defining a curve of constant combined distance around two foci")]
     public class Ellipse : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Origin point defining location of the Ellipse in three-dimensional space")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Defines the major axis of the Ellipse, and the line along which the two foci lie. Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
         public Vector Axis1 { get; set; } = new Vector { X = 1.0, Y = 0.0, Z = 0.0 };
 
+        [Description("Defines the minor axis of the Ellipse. Together with Axis1 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
         public Vector Axis2 { get; set; } = new Vector { X = 0.0, Y = 1.0, Z = 0.0 };
 
+        [Length]
+        [Description("Distance from the Centre to the Ellipse boundary along the major axis")]
         public double Radius1 { get; set; } = 0;
 
+        [Length]
+        [Description("Distance from the Centre to the Ellipse boundary along the minor axis")]
         public double Radius2 { get; set; } = 0;
         
         /***************************************************/
--- a/Geometry_oM/Curve/Ellipse.cs
+++ b/Geometry_oM/Curve/Ellipse.cs
@@ -25,28 +25,28 @@ using System.ComponentModel;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. A standard ellipse defining a curve of constant combined distance around two foci")]
+    [Description("A plane curve. A standard ellipse defining a curve of constant combined distance around two foci.")]
     public class Ellipse : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        [Description("Origin point defining location of the Ellipse in three-dimensional space")]
+        [Description("Origin point defining location of the Ellipse in three-dimensional space.")]
         public Point Centre { get; set; } = new Point();
 
-        [Description("Defines the major axis of the Ellipse, and the line along which the two foci lie. Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
+        [Description("Defines the major axis of the Ellipse, and the line along which the two foci lie. Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
         public Vector Axis1 { get; set; } = new Vector { X = 1.0, Y = 0.0, Z = 0.0 };
 
-        [Description("Defines the minor axis of the Ellipse. Together with Axis1 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
+        [Description("Defines the minor axis of the Ellipse. Together with Axis1 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
         public Vector Axis2 { get; set; } = new Vector { X = 0.0, Y = 1.0, Z = 0.0 };
 
         [Length]
-        [Description("Distance from the Centre to the Ellipse boundary along the major axis")]
+        [Description("Distance from the Centre to a point on the Ellipse along the major axis.")]
         public double Radius1 { get; set; } = 0;
 
         [Length]
-        [Description("Distance from the Centre to the Ellipse boundary along the minor axis")]
+        [Description("Distance from the Centre to a point on the Ellipse along the minor axis.")]
         public double Radius2 { get; set; } = 0;
         
         /***************************************************/
--- a/Geometry_oM/Curve/Ellipse.cs
+++ b/Geometry_oM/Curve/Ellipse.cs
@@ -36,10 +36,10 @@ namespace BH.oM.Geometry
         [Description("Origin point defining location of the Ellipse in three-dimensional space.")]
         public Point Centre { get; set; } = new Point();
 
-        [Description("Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
+        [Description("Together with Axis2 defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
         public Vector Axis1 { get; set; } = new Vector { X = 1.0, Y = 0.0, Z = 0.0 };
 
-        [Description("Together with Axis1 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
+        [Description("Together with Axis1 defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
         public Vector Axis2 { get; set; } = new Vector { X = 0.0, Y = 1.0, Z = 0.0 };
 
         [Length]
--- a/Geometry_oM/Curve/Ellipse.cs
+++ b/Geometry_oM/Curve/Ellipse.cs
@@ -25,7 +25,8 @@ using System.ComponentModel;
 
 namespace BH.oM.Geometry
 {
-    [Description("A plane curve. A standard ellipse defining a curve of constant combined distance around two foci.")]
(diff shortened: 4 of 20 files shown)


al-fisher:
@IsakNaslundBh @pawelbaran - taking this off draft and `wip` ready for review. Not quite all objects covered in Geometry Namespace yet - but keen to get your review and merge - not least of all as I am on holiday 🏖 for a week so do not want to hold up.

Fee free to pick up typos/tweaks committing straight to this branch as needed, and then merge. Cheers 🙏

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,26 +20,31 @@
...
-using System.ComponentModel;
 using System;
-
+using System.ComponentModel;
+using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("Arc starting from the local x, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. Arc starting from the local X, rotating Angle number of radians counter clockwise")]
```
IsakNaslundBh: Reading the preexisting desciption here, it is not fully correct. The Arc starts at local x, rotated 'StartAngle' number of radians counter clockwise about the local z, and ends at `EndAngle` radians. SO might get rid of this all in all as it is being stated on the property?

Also, in Structures_oM we tried using lower case axis letters for local axis and upper case for global axis. Probably want these aligned.

Geometry_oM/CoordinateSystem/Cartesian.cs
```diff
@@ -25,19 +25,23 @@
...
 {
-    [Description("Cartesian coordinate system. All vectors to be orthogonal unit vectors")]
+    [Description("A three-dimensional Cartesian coordinate system, as defined by a set of mutually orthogonal unit vectors and a common origin point")]
     public class Cartesian : IGeometry, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Unit vector defining X coordinate axis")]
```
IsakNaslundBh: As mentioned on the arc, in structures_oM we opted for lower case for local axis. Might be good to align, either this or Structures_oM.

ALso, might be worth spelling out local?

"Unit vector defining local x coordinate axis"

Geometry_oM/Solid/Cuboid.cs
```diff
@@ -21,24 +21,30 @@
...
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Defines both the Cuboid centre as the CoordinateSystem Origin, as well as the orientation in three-dimensional space")]
         public CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new CoordinateSystem.Cartesian();
 
-        [Description("Dimension in the XAxis")]
+        [Length]
+        [Description("Dimension in the X-axis")]
```
IsakNaslundBh: Even though you sort of implicitly can deduce this from the description of the coordinate system, might be good to state how this (and the properties below) relates to the coordinate system. That is, that is (if I read it correctly) that you will have a face Length/2 long x-axis and Length/2 along negative x-axis.

Basically, state that this will work differently than it does for the cylinder/cone.

Also, same upper/lower case comment as before for global/local.

Geometry_oM/Vector/Basis.cs
```diff
@@ -20,25 +20,25 @@
...
 namespace BH.oM.Geometry
 {
+    [Description("An ordered set of vectors defining three-dimensional orientation in space")]
     public class Basis : IGeometry, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("The X coordinate axis")]
```
IsakNaslundBh: lower case for local coordinates (?), and here as well, maybe spell out that it is local (even if it should be quite evident).

Geometry_oM/Curve/Arc.cs
al-fisher: Thanks @IsakNaslundBh - missed that one. Ta! Sounds good. 👍

Geometry_oM/Solid/Cuboid.cs
al-fisher: Yep - useful 👍 thanks

Geometry_oM/CoordinateSystem/Cartesian.cs
al-fisher: 👍

Geometry_oM/Vector/Basis.cs
al-fisher: 👍

al-fisher:
Thanks @IsakNaslundBh

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,27 +20,32 @@
...
 {
-    [Description("Arc starting from the local x, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. A circular Arc, a curve, with a constant distance for a point, its centre, defined it the local angular domain between start and end angle.")]
     public class Arc : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Defines the Arc centre as the CoordinateSystem Origin, as well as the orientation of the Arc in three-dimensional space, lying in the CoordinateSystem XY-Plane")]
```
pawelbaran: Maybe "Defines Arc centre equal to Origin of the CoordinateSystem, as well as the orientation of the Arc in three-dimensional space, equal to local xy-Plane of the CoordinateSystem"?

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,27 +20,32 @@
...
         public CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new CoordinateSystem.Cartesian();
 
+        [Length]
+        [Description("Distance from the CoordinateSystem.Origin to the Arc boundary")]
```
pawelbaran: "Distance from the CoordinateSystem.Origin to any point on Arc"

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,27 +20,32 @@
...
         public double Radius { get; set; } = 0;
 
-        [Description("Angle in radians to the start point from the local X-axis, counter clowise around the local Z-axis")]
+        [Angle]
+        [Description("Angle in radians to the start point from the local x-axis, counter clowise around the local z-axis")]
```
pawelbaran: "Angle in radians to the start point from the local x-axis of CoordinateSystem, counter clockwise around the local z-axis of CoordinateSystem"

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,27 +20,32 @@
...
         public double StartAngle { get; set; } = 0;
 
-        [Description("Angle in radians to the end point from the local X-axis, counter clowise around the local Z-axis")]
+        [Angle]
+        [Description("Angle in radians to the end point from the local x-axis, counter clowise around the local z-axis")]
```
pawelbaran: "Angle in radians to the end point from the local x-axis of CoordinateSystem, counter clowise around the local z-axis of CoordinateSystem"

Geometry_oM/Curve/Circle.cs
```diff
@@ -20,18 +20,26 @@
...
     public class Circle : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Origin point defining location of the Circle in three-dimensional space")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Vector perpendicular to the plane in which the Circle lies in. Defines the orientation in three-dimensional space")]
```
pawelbaran: "Vector perpendicular to the plane in which the Circle lies."

Geometry_oM/Curve/Circle.cs
```diff
@@ -20,18 +20,26 @@
...
         public Vector Normal { get; set; } = new Vector { X = 0, Y = 0, Z = 1 };
 
+        [Length]
+        [Description("Distance from the Centre to the Circle boundary")]
```
pawelbaran: "Distance from the Centre to any point on Circle"

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -20,22 +20,33 @@
...
     public class Ellipse : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Origin point defining location of the Ellipse in three-dimensional space")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Defines the major axis of the Ellipse, and the line along which the two foci lie. Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
```
pawelbaran: Are both foci lying along Axis1, not one of the along Axis1 and the other along Axis2?

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -20,22 +20,33 @@
...
         public Vector Axis1 { get; set; } = new Vector { X = 1.0, Y = 0.0, Z = 0.0 };
 
+        [Description("Defines the minor axis of the Ellipse. Together with Axis1 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters")]
         public Vector Axis2 { get; set; } = new Vector { X = 0.0, Y = 1.0, Z = 0.0 };
 
+        [Length]
+        [Description("Distance from the Centre to the Ellipse boundary along the major axis")]
```
pawelbaran: Does a curve have boundary? Wouldn't it be clearer to say "Distance from the Centre to a point on the Ellipse along the major axis"?

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -20,22 +20,33 @@
...
         public double Radius1 { get; set; } = 0;
 
+        [Length]
+        [Description("Distance from the Centre to the Ellipse boundary along the minor axis")]
```
pawelbaran: Same as above

Geometry_oM/Curve/NurbsCurve.cs
```diff
@@ -20,20 +20,26 @@
...
 {
+    [Description("A non-uniform rational basis spline (or B-spline). Enables definition of smooth and continuous curves as piece-wise polynomial basis functions evaluated from discrete geometrical ControlPoints." +
+                 "\nA NurbsCurve, through its use of ControlPoint Weightings, has the ability to precisely represent the conic sections, such as the circle and parabolas etc.")]
     public class NurbsCurve : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("An ordered set of three-dimensional points defining the curve shape. The control point locations are approximating, as opposed to interpolating, meaning in the general case the resulting curve will not pass through the control point locations themselves")]
```
pawelbaran: Not true for full multiplicity knots?

Geometry_oM/Curve/NurbsCurve.cs
```diff
@@ -20,20 +20,26 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
 using System.Collections.Generic;
 
 namespace BH.oM.Geometry
 {
+    [Description("A non-uniform rational basis spline (or B-spline). Enables definition of smooth and continuous curves as piece-wise polynomial basis functions evaluated from discrete geometrical ControlPoints." +
+                 "\nA NurbsCurve, through its use of ControlPoint Weightings, has the ability to precisely represent the conic sections, such as the circle and parabolas etc.")]
```
pawelbaran: May be worth to mention that we are using the (less mainstream) Rhino approach to Nurbs?

Geometry_oM/Curve/PolyCurve.cs
```diff
@@ -20,16 +20,19 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
 using System.Collections.Generic;
 
 namespace BH.oM.Geometry
 {
+    [Description("A composite curve constructed by combining a collection of curves of any type. Whole PolyCurve integrity, continuity and closure is not guaranteed at creation, thus discontinuous and/or multi-region definitions are possible")]
```
pawelbaran: Worth to mention that although the discontinuous and/or multi-region definitions are possible, they are not recommended as may cause errors in methods operating on PolyCurve?

Geometry_oM/Solid/Cuboid.cs
```diff
@@ -21,24 +21,30 @@
...
         public CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new CoordinateSystem.Cartesian();
 
-        [Description("Dimension in the XAxis")]
+        [Length]
+        [Description("Dimension in the local x-axis. The Cuboid will have one face at a distance of Length/2 from the centre along the positive local x-axis and, one face at a distance of Length/2 from the centre along the negative local x-axis")]
         public double Length { get; set; } = 0.0;
 
-        [Description("Dimension in the YAxis")]
+        [Length]
+        [Description("Dimension in the local y-axis. The Cuboid will have one face at a distance of Depth/2 from the centre along the positive local y-axis and, one face at a distance of Depth/2 from the centre along the negative local y-axis")]
```
pawelbaran: Depth or Width? 🐒

Geometry_oM/Solid/Cylinder.cs
```diff
@@ -34,10 +35,15 @@ public class Cylinder : ISolid
         [Description("Point defining the centre of the circular base")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Vector perperdicular to the base. Defines the orientation in three-dimensional space")]
         public Vector Axis { get; set; } = new Vector { X = 0, Y = 0, Z = 1 };
 
+        [Length]
+        [Description("Distance from the Axis to the Cylinder boundary")]
```
pawelbaran: "Distance from the Axis to the Cylinder boundary surface"

Geometry_oM/Solid/Torus.cs
```diff
@@ -21,22 +21,29 @@
...
 
+        [Description("Vector defining the axis of revolution and the orientation of the Torus in three-dimensional space")]
         public Vector Axis { get; set; } = new Vector { X = 0, Y = 0, Z = 1 };
 
+        [Length]
+        [Description("Distance from the Torus Centre (and the axis of revolution) to the centre of the circular cross section")]
         public double RadiusMajor { get; set; } = 0.0;
 
+        [Length]
+        [Description("The radius defining the size of the circular cross-section")]
```
pawelbaran: "Radius defining distance from the axis of revolution to the boundary surface of the Torus"

Geometry_oM/Vector/Point.cs
```diff
@@ -21,19 +21,24 @@
...
 namespace BH.oM.Geometry
 {
+    [Description("Defines a dimensionless location in three-dimensional space")]
     public class Point : IGeometry, IComparable<Point>, IElement0D
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Position along X coordinate axis")]
```
pawelbaran: "Position along global X coordinate axis"

Geometry_oM/Vector/Point.cs
```diff
@@ -21,19 +21,24 @@
...
         public double X { get; set; } = 0;
 
+        [Description("Position along Y coordinate axis")]
```
pawelbaran: As above

Geometry_oM/Vector/Point.cs
```diff
@@ -21,19 +21,24 @@
...
         public double Y { get; set; } = 0;
 
+        [Description("Position along Z coordinate axis")]
```
pawelbaran: As above

Geometry_oM/Vector/Vector.cs
```diff
@@ -21,19 +21,28 @@
...
 {
+    [Description("Defines a direction and magnitude in three-dimensional Euclidean space")]
     public class Vector : IGeometry, IComparable<Vector>
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Length]
+        [Description("Dimension along X coordinate axis")]
```
pawelbaran: "Dimension along global X coordinate axis"

Geometry_oM/Vector/Vector.cs
```diff
@@ -21,19 +21,28 @@
...
         public double X { get; set; } = 0.0;
 
+        [Length]
+        [Description("Dimension along Y coordinate axis")]
```
pawelbaran: As above

Geometry_oM/Vector/Vector.cs
```diff
@@ -21,19 +21,28 @@
...
         public double Y { get; set; } = 0.0;
 
+        [Length]
+        [Description("Dimension along Z coordinate axis")]
```
pawelbaran: As above

Geometry_oM/Curve/Arc.cs
```diff
@@ -20,27 +20,32 @@
...
-using System.ComponentModel;
 using System;
-
+using System.ComponentModel;
+using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry
 {
-    [Description("Arc starting from the local x, rotating Angle number of radians counter clockwise")]
+    [Description("A plane curve. A circular Arc, a curve, with a constant distance for a point, its centre, defined it the local angular domain between start and end angle.")]
```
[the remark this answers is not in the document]
al-fisher: 👍

Geometry_oM/Curve/Arc.cs
al-fisher: 👍

Geometry_oM/Curve/Arc.cs
al-fisher: 👍

Geometry_oM/Curve/Arc.cs
al-fisher: 👍

Geometry_oM/Curve/Arc.cs
al-fisher: 👍

Geometry_oM/Curve/Circle.cs
al-fisher: 👍

Geometry_oM/Curve/Circle.cs
al-fisher: 👍

Geometry_oM/Curve/Ellipse.cs
al-fisher: Actually nope. 😄  Both foci do lie on the major axis - see https://mathworld.wolfram.com/Ellipse.html
Hope description is OK? Not pushing a change - let me know is OK with it as is? 😄

Geometry_oM/Curve/Ellipse.cs
al-fisher: Yes - agreed! 👍 Making these changes!

Geometry_oM/Curve/NurbsCurve.cs
al-fisher: I might not. If people are _really_ keen will look in code. As long as instructions for inputs etc. are clear would not ref. other implementations

Geometry_oM/Curve/PolyCurve.cs
al-fisher: Oh, yes! Have added this. Good one. Thanks

Geometry_oM/Solid/Cuboid.cs
al-fisher: Hehe! Happy to have an issue raised about renaming Property!! @pawelbaran  Outside of this PRs scope 🐵 !! 😄

Geometry_oM/Curve/NurbsCurve.cs
al-fisher: Hmm. I do not think that is true. A NurbsCurve constructed of full multiplicity knots will certainly go through it's end points. If that is what you meant. But will still interpolate the intermediate control points.
A NurbsCurve with the same number of knots as control points _will_ pass through all control points as a linear interpolation as  is Degree 1. That is why I said "in the general case".
But yeah - I will have a go at adding more info to the Description to clarify.

Geometry_oM/Solid/Torus.cs
al-fisher: Yep - agreed. Changed

Geometry_oM/Solid/Cylinder.cs
al-fisher: Yep - agreed. Changed

Geometry_oM/Vector/Point.cs
al-fisher: 👍

Geometry_oM/Vector/Vector.cs
al-fisher: 👍

Geometry_oM/Vector/Vector.cs
al-fisher: 👍

Geometry_oM/Vector/Vector.cs
al-fisher: 👍

Geometry_oM/Curve/Ellipse.cs
al-fisher: 👍

al-fisher:
Thanks @pawelbaran for your comments - really valuable! I will push some tweaks adopting most of your notes.
A couple of your comments above I have responded to, worth you checking you are happy with my response. 
Thanks 😸 👍

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -20,22 +20,33 @@
...
     public class Ellipse : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Origin point defining location of the Ellipse in three-dimensional space.")]
         public Point Centre { get; set; } = new Point();
 
+        [Description("Defines the major axis of the Ellipse, and the line along which the two foci lie. Together with Axis2 also defines orientation in three-dimensional space. Direction only, and not magnitude, matters.")]
```
IsakNaslundBh: "the line along which the two foci lie."

Is this actually always true for our case? Can easily create an `Ellipse` with `Radius2 > Radius1`, and the Axis2 would be defining the line of the two foci.

al-fisher: Ah right! Actually - good catch.
It is the relative size if our Radii that determines which acis Axis1 or Axis2 is the “Major”!
Thanks - will change descriptions

Geometry_oM/Curve/Ellipse.cs
pawelbaran: It looks like I missed some math classes 🤣 Thanks for the link, all clear now

Geometry_oM/Curve/NurbsCurve.cs
pawelbaran: Oke, looking at it now I think I missed the _general case_ section. Would not complain about it today 😉 LGTM

Geometry_oM/Curve/NurbsCurve.cs
pawelbaran: I did not mean mentioning Rhino explicitly, but rather weaving the information into the description (not even sure if in the class description or maybe one of its properties). Or, alternatively, could be worth to sneak this information into the Wiki. But surely this is not worth delaying this PR.

Geometry_oM/Solid/Cuboid.cs
pawelbaran: #737 raised, as mentioned not in scope of this PR so LGTM :)

al-fisher:
Thanks @pawelbaran - just about to push a final commit. Then perhaps we look to merge if happy. @IsakNaslundBh  
A couple of classes still not annotated but I will do in a second PR, if all happy with that.

al-fisher:
@pawelbaran @IsakNaslundBh I have just tweaked Ellipse and NurbsCurve based on final comments above.

Geometry_oM/Curve/Line.cs
```diff
@@ -20,18 +20,24 @@
...
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Order of Start and End Points defines Line direction, which can be important for some applications")]
         public Point Start { get; set; } = new Point();
 
+        [Description("Order of Start and End Points defines Line direction, which can be important for some applications")]
         public Point End { get; set; } = new Point();
 
+        [Description("A value of true defines the Line as a ray of infinite extents in both directions")]
```
IsakNaslundBh: "A value of true" reads a bit weird in my opinion, but would not block the PR for this reason.

Geometry_oM/Curve/Line.cs
```diff
@@ -20,18 +20,24 @@
...
 namespace BH.oM.Geometry
 {
+    [Description("A straight segment in space defining the shortest distance between two points in three-dimensional Euclidean geometry")]
     public class Line : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Order of Start and End Points defines Line direction, which can be important for some applications")]
```
IsakNaslundBh: I get what you want to say with this comment, and why it sits here, but feels a little bit strange to have this general description on the Start/End. But might be ok. Maybe could be rephrased for the two, to focus in on their particular property a bit more somehow. Same as comment below, will not let it hold up the PR.

Geometry_oM/Curve/NurbsCurve.cs
```diff
@@ -20,20 +20,29 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
 using System.Collections.Generic;
 
 namespace BH.oM.Geometry
 {
+    [Description("A non-uniform rational basis spline (or B-spline). Enables definition of smooth and continuous curves as piece-wise polynomial basis functions evaluated from discrete geometrical ControlPoints." +
```
IsakNaslundBh: Details, but smooth only guarantied for degree > 1?

Geometry_oM/Vector/Basis.cs
```diff
@@ -20,25 +20,25 @@
...
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using BH.oM.Base;
+using System.ComponentModel;
 
 namespace BH.oM.Geometry
 {
+    [Description("An ordered set of vectors defining three-dimensional orientation in space")]
```
IsakNaslundBh: Maybe add, similar to the Cartesian coordinate system, that the vectors should be mutually orthogonal unit vectors? At least, that is assumed for many cases. This is unless you intentionally left this more open to allowed for various skewed basii (is that the plural form?) as well.

Geometry_oM/Vector/Basis.cs
```diff
@@ -20,25 +20,25 @@
...
     public class Basis : IGeometry, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("The local x coordinate axis")]
```
IsakNaslundBh: If comment above holds true, this should say that these are unit vectors as well.

Geometry_oM/Vector/Plane.cs
```diff
@@ -20,16 +20,21 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
+
 namespace BH.oM.Geometry
 {
+    [Description("Defines a flat infinite two-dimensional surface")]
```
IsakNaslundBh: Risk of mixing up the word 'surface' here with out other general surfaces? After all, the Plane does not (and should not imo) implement the `ISurface` interface. Maybe simply 

`Defines a infinite two-dimensional plane in three-dimensional space`

or something could work?

Geometry_oM/Vector/Plane.cs
```diff
@@ -20,16 +20,21 @@
...
     public class Plane : IGeometry
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("A point through which the plane passes")]
         public Point Origin { get; set; } = new Point();
 
+        [Description("Vector perpendicualar to the plane, defining the orientation of the surface in three-dimensional space")]
```
IsakNaslundBh: see comment above regarding the use of the word 'surface'

Geometry_oM/Vector/Point.cs
```diff
@@ -21,20 +21,26 @@
  */
 
 using System;
+using System.ComponentModel;
 using BH.oM.Dimensional;
 
+
 namespace BH.oM.Geometry
 {
+    [Description("Defines a dimensionless location in three-dimensional space.")]
```
IsakNaslundBh: not sure i get 'dimensionless' ?

Geometry_oM/Vector/Plane.cs
al-fisher: hmmm... trying to avoid circular definition - i.e. not stating plane in the definition of plane.
I actually do not have a problem with "surface" here. Crucially we are stating is _infinite_ surface. Then actually your point on risk of mixing up with other Surfaces - I might actually disagree. Plane _could_ implement `ISurface`, but one for later.

Geometry_oM/Vector/Plane.cs
al-fisher: will rephrase anyway as redundant

Geometry_oM/Vector/Point.cs
al-fisher: In that it has no size.

Geometry_oM/Vector/Basis.cs
al-fisher: Yep! Good one. Thanks for the catch. 👍 
Will match Cartesian. I was not intending to keep general for skewed - think we should pick that up separately

Geometry_oM/Curve/Line.cs
al-fisher: Can reword.

Geometry_oM/Curve/NurbsCurve.cs
al-fisher: Enables, but does not guarantee. 😉

Geometry_oM/Vector/Basis.cs
al-fisher: yep! Thanks

al-fisher:
> "in space"
"in three-dimensional space"
"in three-dimensional Euclidean space"
Might be good to align the wording to use the same everywhere, unless this was intentional.

It was intentional, and was well meaning, but if confusing could be changed. To be honest I might leave as is now as is not _wrong_. We will have plenty of opportunity to continue to tweak when rendered into API. THIS WILL NEVER END! 😄 😆

Geometry_oM/Curve/Line.cs
al-fisher: Hmmm ...  yes - have moved this point up to the line's general description.

al-fisher:
Thanks @IsakNaslundBh - have incorporated some of your comments.

Geometry_oM/Curve/NurbsCurve.cs
IsakNaslundBh: true! 😄 Then fine, ignore the comment! 🛩

Geometry_oM/Vector/Plane.cs
IsakNaslundBh: Yeah, get your point there. Tried thinking of another synonym, but drew a bit blank..

And yeah, if a plane is a surface or not most def sounds like something better handled outside this PR 😄

Geometry_oM/Vector/Point.cs
IsakNaslundBh: ah, ofc. Should have understood that from all our discussions regarding this! 😉 

But as I did not straight away, might be better off without the word? Saying that. It is a Point. I think most people will understand what a Point is! 😄

Geometry_oM/Vector/Plane.cs
al-fisher: The correct general word is manifold - but think surface is clearer

Geometry_oM/Curve/Line.cs
```diff
@@ -20,18 +20,25 @@
...
 {
+    [Description("A straight segment in space defining the shortest distance between two points in three-dimensional Euclidean geometry.\n" +
+                 "The Vector from Start to End defines the Line direction, which can be important for some applications.")]
     public class Line : ICurve
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Position in three-dimensional space.")]
```
IsakNaslundBh: Would consider removing the description all in all. The text here will get picked up from the Point description anyway. Right now it becomes sort of a duplication:

image

Geometry_oM/Vector/Point.cs
al-fisher: I think as people know what a point is - again I personally think is not a big issue. 
The reason for keeping is it references our element0D interface. And part of these descriptions role is to state what we mean by certain objects.
A point is 0D

Geometry_oM/Curve/Line.cs
al-fisher: Good idea.

al-fisher: Have removed for now at least @IsakNaslundBh

al-fisher: We might need to revisit the format of these composite descriptions as part of further refinement of reflecting into API anyway.
Descriptions may need to be slightly different for API vs UI.
We will have to see
@IsakNaslundBh @FraserGreenroyd FYI

PR #704, opened 2020-02-14
Dimensional_oM: Migrate IElementXD interfaces from Geometry_oM to Dimensional_oM

kThorsager:
### Issues addressed by this PR

Closes #698 
This change will break any project using any object inheriting from `IElement`

### Changelog
- Also adds appropriate references in other oM's
- Correct dependencies for Dimensional_oM

PR #711, opened 2020-02-22
BHoM: Rename IBHoMFragment -> IFragment

### LINKED ISSUE #651: Rename IBHoMFragment as IFragment
As discussed https://github.com/BHoM/BHoM/issues/646#issuecomment-573802526 requirement to rename `IBHoMFragment` and `IFragment`

The fragment interface implements only `IObject` and thus agreed naming convention suggests it should not have "`BHoM`" in the class/interface name.

FraserGreenroyd:
Fixes #651

### DIFF
--- a/BHoM/BHoM.csproj
+++ b/BHoM/BHoM.csproj
@@ -59,7 +59,7 @@
   <ItemGroup>
     <Compile Include="CustomObject.cs" />
     <Compile Include="FragmentSet.cs" />
-    <Compile Include="Interface\IBHoMFragment.cs" />
+    <Compile Include="Interface\IFragment.cs" />
     <Compile Include="Interface\IBHoMGroup.cs" />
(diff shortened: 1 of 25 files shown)


FraserGreenroyd:
Ok looks like there were not many toolkits which needed to update based on this from the analysis done, and all have now been done and set as dependent on this PR.

I do not believe this will clash with #704 but am advising we merge AFTER #704 regardless due to the timeline of planned work.

However, it would be good to have all reviewed before the #704 merge party so we could potentially do both in the same party if everyone is happy.

Also, in hindsight, should have left this to @IsakNaslundBh as it turns out most toolkits (well, all outside the core) impacted were structures related, but this allows @IsakNaslundBh to be able to test and be happy the change does not adversely affect his toolkits, so pros and cons I guess...

al-fisher:
Thanks @FraserGreenroyd - above sounds like a plan

PR #742, opened 2020-03-10
Architecture_oM: Update room to be IElement2D

### LINKED ISSUE #741: Architecture_oM: Add IElement2D interface to room
As discussed with @BingWangUS, we need to add iElement2D to the 'BH.oM.Architecture.Elements.Room' object, to be able to use model laundry methods directly to room object instead of translating room object to panel as current workflow.

FraserGreenroyd:
Fixes #741

### DIFF
--- a/Architecture_oM/Elements/Room.cs
+++ b/Architecture_oM/Elements/Room.cs
@@ -26,7 +26,7 @@ using BH.oM.Dimensional;
 
 namespace BH.oM.Architecture.Elements
 {
-    public class Room : BHoMObject, IElement1D
+    public class Room : BHoMObject, IElement2D
     {
         /***************************************************/
         /**** Properties                                ****/


al-fisher:
> This does what the issue wants it to do, but won't it break every script? Not that I'm sure on how to handle that properly (can't be both interfaces) and might not be something which needs to be considered

Responding to your comment @kThorsager - the Room was not implementing IElement1D properly anyway - in fact it could not reasonably be considered 1D. So think fine to make the simple change as proposed. Will merge now

BingWangUS:
Thanks for responding to this quickly. I tested the code after merged, throwing me the following error when trying to apply SnapToGrids (2d) method.

1. This component failed to run properly.
- Error: Value cannot be null.
Parameter name: source
- Occured in  BH.Engine.ModelLaundry.Modify.SnapToGrids(IElement2D element2D, List`1 grids, Double maxDistanceExternal, Double maxAngleExternal, Double maxDistanceInternal, Double maxAngleInternal, Boolean allowNonPlanar, Double tolerance)
     called from BH.Engine.ModelLaundry.Modify.SnapToGrids(IElement2D element2D, List`1 grids, Double maxDistance, Double maxAngle, Boolean allowNonPlanar, Double tolerance)
- Are you sure you have the correct type of inputs? Check their description for more details.

PR #748, opened 2020-03-16
LifeCycleAssessment_oM: Migration of LifeCycleAssessment objects

### LINKED ISSUE #744: LifeCycleAssessment_oM : Migrate LCA Objects to BHoM
#### Description:
Migrate objects currently in LifeCycleAssessment_Toolkit to remove dependencies for CarbonQueryDatabase_Toolkit and others.

michaelhoehn:
+Update Code owners file.

   
Closes #744


LifeCycleAssessment_oM/EnvironmentalProductDeclaration.cs
```diff
@@ -1,12 +1,63 @@
...
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+        public string Manufacturer { get; set; } = "";
+        public string Plant { get; set; } = "";
+        public int PostalCode { get; set; } = 0;
+        public List<string> IndustryStandards { get; set; } = new List<string>();
+        public string Id { get; set; } = "";
+        public string Density { get; set; } = "";
+        public string GwpPerKG { get; set; } = "";
```
enarhi: Lets get rid of these two GWP properties as GlobalWarmingPotential and DeclaredUnit render them redundant. We will need to track this through with EvaluateEPD and the convert methods in LCA and CQD toolkit.

LifeCycleAssessment_oM/SectorEnvironmentalProductDeclaration.cs
```diff
@@ -1,12 +1,58 @@
...
+    public class SectorEnvironmentalProductDeclaration : BHoMObject, IEnvironmentalProductDeclarationData
     {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+        public List<string> Publisher { get; set; } = new List<string>();
+        public List<string> Jurisdiction { get; set; } = new List<string>();
+        public string Id { get; set; } = "";
+        public string Density { get; set; } = "";
+        public string GwpPerKG { get; set; } = "";
```
enarhi: Eliminate redundant GWP props as above.

LifeCycleAssessment_oM/HealthProductDeclaration.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace BH.oM.LifeCycleAssessment
+{
+    public class HealthProductDeclaration : BHoMObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+        //public string Cpid { get; set; } = "";
+        //public string Version { get; set; } = "";
```
FraserGreenroyd: If these aren't being used let's delete them

LifeCycleAssessment_oM/IEnvironmentalProductDeclarationData.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace BH.oM.LifeCycleAssessment
+{
+    public interface IEnvironmentalProductDeclarationData : IBHoMObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+        string Id { get; set; }
+        string Density { get; set; }
+        string DeclaredUnit { get; set; } // <---- make sure this is always populated
```
FraserGreenroyd: This comment probably isn't needed

LifeCycleAssessment_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("LifeCycleAssessment_oM")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("LifeCycleAssessment_oM")]
+[assembly: AssemblyCopyright("Copyright ©  2020")]
```
FraserGreenroyd: Copyright should be [assembly: AssemblyCopyright("Copyright © https://github.com/BHoM")]

LifeCycleAssessment_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
FraserGreenroyd: Assembly Version should be `3.0.0.0`

LifeCycleAssessment_oM/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
FraserGreenroyd: Assembly File Version should be `3.1.0.0` but realistically this won't be merged until after the end of the 3.1 milestone so perhaps best to set them to `3.2.0.0`

FraserGreenroyd:
@al-fisher it would be good to get your review on this in the next few days to begin merging this out and consolidate the workflows for LifeCycleAssessment_Toolkit and CarbonQueryDatabase_Toolkit for Installer inclusion if possible :smile:

LifeCycleAssessment_oM/IEnvironmentalProductDeclarationData.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace BH.oM.LifeCycleAssessment
+{
+    public interface IEnvironmentalProductDeclarationData : IBHoMObject
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+        string Id { get; set; }
+        string Density { get; set; }
```
IsakNaslundBh, line 34: Density as a string looks a bit strange to me? Any reason this can not be a double?

LifeCycleAssessment_oM/IEnvironmentalProductDeclarationData.cs
```diff
@@ -0,0 +1,54 @@
...
+        string DeclaredUnit { get; set; }
```
IsakNaslundBh, line 35: Know this have been discussed, and understand the reasons for why this sits here, but would again slightly challenge if it is absolutely needed. We have been able to get away without this up til this point, would if at all possible prefer the same here.

If not, would it be possible to change this to dimensions, rather than unit. Basically something that states "PerArea" or "PerVolume" or "PerItem" etc, rather than specifying the exact unit (`[ft^3]`/`[m^2]` etc), and then try to keep the units used in SI, consistent with the rest of the BHoM.

Could then potentially add an enum or similar for the options available.

@al-fisher

al-fisher: > If not, would it be possible to change this to dimensions, rather than unit. Basically something that states "PerArea" or "PerVolume" or "PerItem" etc, rather than specifying the exact unit ([ft^3]/[m^2] etc), and then try to keep the units used in SI, consistent with the rest of the BHoM.

+1 on this idea.
Can do as a proper `QuantityAttribute` then also 👍

al-fisher: Will enable type strong checking that you are not wrongly multiply a "per Area" by a volume etc.

enarhi: @al-fisher @IsakNaslundBh the issue with this is that the declared unit is not always 1. It could be per 5 pieces, per 0.76455 m3, per 1000kg, etc. Because of this we either need dimension and value, or separate properties (ie DeclaredUnitType and DeclaredUnitQuantity). Would it make more sense to do it that way, or is there an alternate option I'm missing that could cover both in one go? I think the one thing we want to make sure we avoid is excessive need for any epd related toolkit / dataset to handle unit and string conversions itself.

LifeCycleAssessment_oM/SectorEnvironmentalProductDeclaration.cs
```diff
@@ -0,0 +1,56 @@
...
+        public double AcidificationPotential { get; set; } = double.NaN;
+        public double EutrophicationPotential { get; set; } = double.NaN;
+        public double DepletionOfAbioticResourcesFossilFuels { get; set; } = double.NaN;
+        public double GlobalWarmingPotentialEndOfLife { get; set; } = double.NaN;
+        public double OzoneDepletionPotentialEndOfLife { get; set; } = double.NaN;
+        public double PhotochemicalOzoneCreationPotentialEndOfLife { get; set; } = double.NaN;
+        public double AcidificationPotentialEndOfLife { get; set; } = double.NaN;
+        public double EutrophicationPotentialEndOfLife { get; set; } = double.NaN;
+        public double DepletionOfAbioticResourcesFossilFuelsEndOfLife { get; set; } = double.NaN;
+        public string EndOfLifeTreatment { get; set; } = "";
```
FraserGreenroyd, line 53: Apologies @michaelhoehn , since the merging of the `virtual` PR can we add the `virtual` keyword to all properties on this oM :smile:

LifeCycleAssessment_oM/IEnvironmentalProductDeclarationData.cs
IsakNaslundBh: > @al-fisher @IsakNaslundBh the issue with this is that the declared unit is not always 1. It could be per 5 pieces, per 0.76455 m3, per 1000kg, etc.

For those cases I would aim to scale the values at creation of the object, to match. like
- For 5 peices, state "PerItem" and divide all values by 5
- For per 0.76455 m3, state per volume, and divide all the values by 0.76455
- For per 1000kg, state per mass and divide all values by 1000 etc.

That way you get everything aligned.

> ? I think the one thing we want to make sure we avoid is excessive need for any epd related toolkit / dataset to handle unit and string conversions itself.

I agree on this fully, and that is the main reason I would think that it is better if any data serialised out can be assumed to be in a uniform unit system, with a limited set of options on what dimensions to cover. PerValume always means per m3, per mass always means per kg, etc. That way, you will end up with a very limited set of operations you need to write when working with the data later. Basically, to avoid having to deal with the case that someone has said "per 0.2463 cm^2" etc.

> Would it make more sense to do it that way, or is there an alternate option I'm missing that could cover both in one go?

Hoping the above could. Unit being implicit as SI as it is across the whole of BHoM, and quantity being explicit by the enum.

I know this puts a bit more effort on people generating the data, but once that has been done, it should make it a whole lot easier to make use of it. Also think the person generating the data will be far better equipped to make sure the data comes in correctly, compared to a user getting it and themselves having to make sense of some units in a string etc.

enarhi:
Also, if we are planning on keeping LifeCycleAssessment_oM in the LCA toolkit as well, we will have a clash with the BHoM LCA_oM DLL and the LCA_Toolkit LCA_oM dll. Should we rename these to something other than LifeCycleAssessment as long as we are planning on having other objects defined in the LCA_Toolkit's LCA_oM?

FraserGreenroyd:
> Also, if we are planning on keeping LifeCycleAssessment_oM in the LCA toolkit as well, we will have a clash with the BHoM LCA_oM DLL and the LCA_Toolkit LCA_oM dll. Should we rename these to something other than LifeCycleAssessment as long as we are planning on having other objects defined in the LCA_Toolkit's LCA_oM? @FraserGreenroyd

Providing the namespaces are not exact matches then this shouldn't be an issue. This is what we've done with other toolkits and 'injecting' into existing namespaces. For example, TriangleNet_Toolkit and Clipper_Toolkit both inject into `BH.oM.Geometry` by giving a new subnamespace which becomes unique and does not cause clashes between the DLLs,

As it currently stands, there would be a clash but I am imagining that most of the objects being created withing LifeCycleAssessment_Toolkit will migrate over to here as well - @michaelhoehn is that the plan?

enarhi:
@FraserGreenroyd makes sense. So that means we should move the the LifeCycleAssessment_Toolkit om work here and do all om work going forward here instead of in the LCA_toolkit since they are identical namespaces, correct?

FraserGreenroyd:
It would be my recommendation yes. Perhaps something for us to discuss in our meeting next week. It might be worth @michaelhoehn looking at moving the objects which have just gone into LifeCycleAssessment_Toolkit to here as part of discussions, or the namespace on that toolkit should be subnamespaced out to help avoid conflicts with CarbonQueryDatabase_Toolkit for the interim.

al-fisher:
Yes - I would avoid splitting the LifeCycleAssessment_oM across multiple dlls and repos if at all possible. Will be more logical for finding the code and developing also.
Splitting makes sense when mirroring a very logical scope split- and in that case I would also suggest both the namespaces and also the actual .dlls therefore are also clearly named to reflect the scope. Keeping intuitive

michaelhoehn:
> As it currently stands, there would be a clash but I am imagining that most of the objects being created withing LifeCycleAssessment_Toolkit will migrate over to here as well - @michaelhoehn is that the plan?

@FraserGreenroyd yes that is the plan. I'll go ahead and migrate all the latest objects over now, and we can move forward doing all oM work within the BHoM LCA_oM to avoid further conflicts. Thanks all!

michaelhoehn:
@FraserGreenroyd @IsakNaslundBh @al-fisher @enarhi 
All objects have been migrated from the LCA_toolkit. Would suggest we push to merge this PR as part of our discussion taking place tomorrow (20201404) and we can open subsequent issues/PRs addressing any further ideas as this branch is already getting a bit stale. I would like to integrate things like IElementM and begin testing as soon as possible.

enarhi:
Couple additional commits to ensure full oM migration and clean build. Should be good to go know, @FraserGreenroyd please review.

FraserGreenroyd:
> Couple additional commits to ensure full oM migration and clean build. Should be good to go know, @FraserGreenroyd please review.

I'll get on this tomorrow as a first task

PR #753, opened 2020-03-20
Environment_oM: Create Exhaust Object

### LINKED ISSUE #752: Environments_oM: Add Exhaust Object
#### Description:
Exhaust object to be created similar to the ventilation object. Properties to include: UnitContinuous, UnitIntermittent, SpaceTypeContinuous

kayleighhoude:

Closes #752 

### Test files

image

### Changelog
Added exhaust object

### DIFF
--- a/Environment_oM/Gains/Exhaust.cs
+++ b/Environment_oM/Gains/Exhaust.cs
@@ -1,0 +1,47 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using System.ComponentModel;
+
+namespace BH.oM.Environment.Gains
+{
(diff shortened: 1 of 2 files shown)


cskinnerBH:
Having a bit of trouble reviewing this one too - fetched the branch, rebuilt the solution, but the new object isn't showing up for me even though it's in the folder structure.

Environment_oM/Gains/Exhaust.cs
```diff
@@ -0,0 +1,47 @@
...
+using BH.oM.Base;
+using System.ComponentModel;
+
+namespace BH.oM.Environment.Gains
+{
+    [Description("Exhaust is indicative of the amount of air that should be extracted from a space based on the space type (eg laboratory, woodshop) or fixtures contained within the space (eg toilets, showers)")]
+    public class Exhaust : BHoMObject
+    {
+        [Description("The continuous exhaust rate associated with a particular space type (eg pet shop, printing room) measured in m3/s*m2")]
+        public double SpaceTypeContinuous { get; set; } = 0.0;
```
FraserGreenroyd: I'm not convinced I like the name of this variable. Can it be called `ContinuousExhaustRate` or similar? I also feel that the measurement `m3/s*m2` should be a quantities attribute rather than given as part of the description.

kayleighhoude: @FraserGreenroyd `ContinuousExhaustRate` may be ambiguous in this instance, there are 3 potential exhaust rate methods (2 of which are continuous): continuous for a room as a whole (woodshop), by fixture which is switched (toilet exhaust connected to a light switch), or by fixture continuous (a constant low flow toilet exhaust). 

For some of the BEnv objects we've listed the units where they may seem ambiguous, no one is used to m3 for airflow (not even the metric folks!) so it's important to state it.

FraserGreenroyd: Ok that makes sense, I'm happy with that for the time being then :smile:

kayleighhoude:
@cskinnerBH i included the csproj file last night, you should be able to rebuild no problem today. once you review we're all set!

cskinnerBH:
I don't see anything inherently wrong in the actual code but when I rebuild I get the following error putting the Exhaust object down. Possibly rebuilt in the incorrect order? Or, missing something I should have rebuilt?

image

FraserGreenroyd:
That does sound like a build error @cskinnerBH - though not sure why. But that object is only referring to primitive types, so that error is likely from the UI caused by the wrong build I would think. Certainly I did not get the same.

Are you happy at least with the objects properties and descriptions?

Environment_oM/Gains/Exhaust.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        [Description("The exhaust rate associated with a particular fixture type (eg toilets, showers) that is run continuously, meaning that it is not triggered by a sensor or switch and is measured in m3/s*unit")]
+        public double UnitContinuous { get; set; } = 0.0;
+
+        [Description("The exhaust rate associated with a particular fixture type (eg toilets, showers) that is run intermittently, meaning that it is triggered by a sensor or switch and is measured in m3/s*unit")]
+        public double UnitIntermittent { get; set; } = 0.0;
```
FraserGreenroyd, line 44: Now that the `virtual` keyword work has been merged, these properties also require that keyword. Are you happy to add the keyword @kayleighhoude or would you like me to do it on this occasion?

FraserGreenroyd:
@al-fisher / @adecler  this is ready for merge :smile:

PR #754, opened 2020-03-23
Geometry_oM: Updates to Object Descriptions

### LINKED ISSUE #700: Geometry_oM: Addition of description attributes to Geometry_oM

al-fisher:
Closes #700


Geometry_oM/SettingOut/Level.cs
```diff
@@ -20,18 +20,25 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
 using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Geometry.SettingOut
 {
+    [Description("Defines a plane, or infinite flat two-dimensional surface, based on a height measured along the global Z axis and origin.")]
```
IsakNaslundBh:
```suggestion
    [Description("Defines a horizontal plane, or infinite flat two-dimensional surface, based on a height measured along the global Z axis and origin.")]
```

Geometry_oM/Surface/Extrusion.cs
```diff
@@ -20,18 +20,24 @@
...
 namespace BH.oM.Geometry
 {
+    [Description("A constant cross section surface, defined by a single profile curve and a linear axis.")]
     public class Extrusion : ISurface
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("The Extrusion profile curve. Slicing the surface at any point perpendicular to the Direction axis Vector will result in an identical curve.")]
```
IsakNaslundBh: Maybe mention something about this preferably being a planar curve.

Then, we are not forcing orthogonality between the curve plane and the vector, hence, not sure this comment is true? Saying that, that renders my comment a bit wrong as well.

Saying all of this, we could start enforcing all of this by making it immutable, and doing checks in the create methods.

Geometry_oM/Surface/ISurface.cs
```diff
@@ -20,8 +20,11 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
+
 namespace BH.oM.Geometry
 {
+    [Description("The interface for all two-dimensional geometrical objects.")]
```
IsakNaslundBh: Mesh does not implement this interface, so might need to narrow this down a little bit.

Geometry_oM/Surface/Pipe.cs
```diff
@@ -20,18 +20,24 @@
...
     public class Pipe : ISurface
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Defines the central axis perpendicular to the circular cross sections, and thus the path along which the tubular surface follows.")]
         public ICurve Centreline { get; set; } = new Line();
 
+        [Description("The distance from the Curve Centreline to a point on the Pipe surface.")]
```
IsakNaslundBh: `[Length]` attribute

Geometry_oM/Surface/Extrusion.cs
```diff
@@ -20,18 +20,24 @@
...
         public ICurve Curve { get; set; } = new Line();
 
+        [Description("The axis along which to translate the profile curve. The Vector magnitude determining the Extrusion length.")]
```
IsakNaslundBh: Maybe `[Length]` attribute?

Geometry_oM/Surface/ISurface.cs
al-fisher: As discussed Mesh is more general and can represent solid and non-manifold as well as surfaces. Point still stands. will remove the "all" for the time being.

Geometry_oM/Surface/Extrusion.cs
al-fisher: Hmmm... good points!

al-fisher: Replacing with this 
``` C#
[Description("The Extrusion profile curve defining the constant cross sectional shape. Idential curves can be recreated at any point along the Extrusion axis through intersection with the Surface.")]`
```

al-fisher: 😄

al-fisher:
Thanks @IsakNaslundBh - have addressed comments above and added final NurbsSurface Descriptions.

Geometry_oM/Surface/PolySurface.cs
```diff
@@ -20,16 +20,19 @@
...
 namespace BH.oM.Geometry
 {
+    [Description("A composite surface constructed by combining a collection of surfaces of any type. Whole PolySurface integrity, continuity and closure is not guaranteed at creation. Discontinuous and/or multi-region definitions are possible, although validity of some engine methods may vary depending on the PolySurface form.")]
     public class PolySurface : ISurface
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("A collection of surfaces, of any or mixed type, which together define the composite shape.")]
```
IsakNaslundBh: details now, but 'composite shape', was this intended, or copied over from composite geometry?

al-fisher: Was written first actually! Copied the other way! 😄 haha!
A PolySurface is a composite shape - made up of multiple surfaces? Does that not sound right.

al-fisher:
Thanks @IsakNaslundBh 🎉

PR #770, opened 2020-03-31
Structure_oM: Add IProperty interface to all property type objects

### LINKED ISSUE #486: Structure_oM: Add IProperty interface
This would make it easier in the converts where some applications have one type of constraint (6DOF) for all/most use cases. @IsakNaslundBh thoughts?

IsakNaslundBh:

Closes #486 

Adding `IProperty` interface to all structural Property type objects.

This is to enable https://github.com/BHoM/BHoM_Engine/issues/1398

This is only additive, so should not break anything.

### Test files

No should be needed.

### DIFF
--- a/Structure_oM/Constraints/BarRelease.cs
+++ b/Structure_oM/Constraints/BarRelease.cs
@@ -27,7 +27,7 @@ namespace BH.oM.Structure.Constraints
 {
 
     [Description("Defines the releases at the start and end node of a Bar.")]
-    public class BarRelease : BHoMObject
+    public class BarRelease : BHoMObject, IProperty
     {
         /***************************************************/
         /**** Properties                                ****/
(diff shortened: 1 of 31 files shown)


IsakNaslundBh:
> Is there a reason `IIsotropic `and `IOrthotropic `don't inherit `IProperty `- I appreciate they inherit from `IMaterialFragment `which inherits `IProperty `but I thought we were inheriting from `IProperty `on all classes as it had no adverse effects and made it clearer.

I added it to the base interface needing it, as well as to the actual classes. Opted to not add it to any intermediate interface, as not needed, but happy to add it in there if you think it is useful?

IsakNaslundBh:
After chat with @al-fisher removed `IProperty` interface from classes that gets it via another interface

PR #776, opened 2020-04-02
Diffing_oM: Corrections to the HashFragment and DiffConfig

### LINKED ISSUE #774: Diffing_oM: HashFragment should not have a constructor
#### Broken rules:
Even if it is not a BHoM object, HashFragment should not have a constructor. This basically prevent reflection and serialisation to work properly with that type of object.

Delete constructor

alelom:

Closes #774 

### Test files

Not required. To be safe, I tested anyway the workflow script:
https://burohappold.sharepoint.com/:f:/s/BHoM/EhZG8bcBYJJAvU00CCGa8AwBVxDOnuqe_vJiTp6v_e5W8g?e=Qeletj
and it works fine.

### Changelog

- `HashFragment` now correctly implements `IImmutable` and its properties are get-only
- `HashFragment` property `Hash` renamed to `CurrentHash`
- Removed legacy properties from `HashFragment` (Name, BHoM_Guid, etc.) that had to be there for the old `IBHoMFragment` implementation to work.
- Added descriptions in `HashFragment`.
- `DiffConfig`: removed `Fragments` from the `PropertiesToIgnore`. This adds the ability to diff also the Fragments; `HashFragment` is ignored automatically in the Engine (see [related PR](https://github.com/BHoM/BHoM_Engine/pull/1640)).


Diffing_oM/HashFragment.cs
```diff
@@ -23,25 +23,30 @@
...
+using System.ComponentModel;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
 namespace BH.oM.Diffing
 {
-    public class HashFragment : IFragment
+    [Description("Stores the current object Hash and its previous Hash.")]
+    public class HashFragment : IFragment, IImmutable
```
alelom, line 34: History/State/Versioning/Revision
@al-fisher says History

Diffing_oM/HashFragment.cs
```diff
@@ -23,25 +23,30 @@
...
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        public string Hash { get; set; }
-        public string PreviousHash { get; set; } = null; // Until we solve the 'Persistent GUID', this is the current workaround.
+        [Description("Current object Hash.")]
+        public string Hash { get; }
```
alelom, line 41: CurrentHash

alelom:
After https://github.com/BHoM/BHoM/pull/776/commits/168a85e58e29e7334535f0d3ffb0c2abced9d366
getting this weird error:
image

will revert that commit and separate that in another PR.

adecler:
It seems to me that you have tagged most of the objects in the Diffing oM as `IImmutable` even when all the properties are orthogonal to each other. Don't forget the purpose of `IImmutable` is to make sure we never have invalid version of an object (i.e. incompatible property values), not because you'd rather not have the user modifying it after construction (for whatever reason).

alelom:
@al-fisher I've applied the most recent corrections as discussed and tested thoroughly. 
This is ready for you to review. Thanks

Diffing_oM/Delta.cs
```diff
@@ -47,28 +47,31 @@ public class Delta : IObject, IImmutable
...
+        public string Author { get; }
 
         public string Comment { get; }
 
         /***************************************************/
         /**** Constructor                               ****/
         /***************************************************/
 
-        public Delta(Guid streamId, Diff diff, Guid revision_from, Guid? revision_to = null, string comment = null)
+        public Delta(Guid streamId, Diff diff, Guid revision_from, Guid revision_to, long timestamp, string author, string comment = null)
```
adecler, line 64: I would keep the default values here:
- `revision_to = null ? Guid.Empty`
- `timestamp = 0 ? DateTime.UtcNow.Ticks`
- `author = "" ? Environment.UserDomainName + "/" + Environment.UserName`

In case my notation doesn't make sense, the `?` is to indicate what you would set the property to if the corresponding input value is at default value.

Serialisation needs all orthogonal properties to have a corresponding argument in the constructor but those argument are still allowed to have default value.

alelom:
Thank you very much @al-fisher ! Sorry I had missed the notification about this!

You are absolutely right on both points, thanks a lot for highlighting!

#### Point 1
Point 1. is now solved by https://github.com/BHoM/BHoM/pull/776/commits/3c82930331d30c61f8edf292be7e182a65affb80 as agreed during the call. This corrects the constructor to be as per compliance for the serialization/deserialisation so it takes all readonly properties.

#### Point 2
Point 2, again completely right. I should have highlighted this before.
In the call I mentioned this is due to a current limitation of the Explode component, but I'm having a new look at this and I think I have changed my mind. It's more because there is a mismatch between what the Explode component rightly does and what should be the intended behaviour from an user perspective. 
This, in theory, calls for another oM type to be introduced for the `ModifiedPropertyPerObject` field, but I am not sure I'd like to introduce another "Diff-type" object.
Also, from a coding perspective I still think the data type make sense as it is now – even though it has admittedly a convoluted form (`Dictionary<string, Dictionary<string, Tuple<string, object>`), [whose necessity is explained also in its description](https://github.com/BHoM/BHoM/blob/a7ca36115a226e19d36247e80412c4bcdcdbc496/Diffing_oM/Diff.cs#L49).

Anyway, I will now raise an issue about this. I might take some thinking about how this should be best tackled. I still think this should be treated separately and this should be good to go with this last corrections.

For the record, here is the example script for the "correct explosion" of the `ModifiedPropertyPerObject`:
<details>

```cs

private void RunScript(object x, ref object objsHash, ref object propName, ref object propValue_Current, ref object propValue_Read)
  {
    var objsHashTree = new Grasshopper.DataTree<string>();
    var propNameTree = new Grasshopper.DataTree<string>();
    var propValue_CurrentTree = new Grasshopper.DataTree<object>();
    var propValue_ReadTree = new Grasshopper.DataTree<object>();

    var dict = new Dictionary<string, Dictionary<string, Tuple<object, object>>>();
    dict = (Dictionary<string, Dictionary<string, Tuple<object, object>>>) x;

    int i = 0;
    foreach (var item in dict)
    {
      var path = new GH_Path(i);

      objsHashTree.Add(item.Key, path);

      foreach (var propItem in item.Value)
      {
        propNameTree.Add(propItem.Key, path);
        propValue_CurrentTree.Add(propItem.Value.Item1, path);
        propValue_ReadTree.Add(propItem.Value.Item2, path);
      }

      i += 1;
    }

    objsHash = objsHashTree;
    propName = propNameTree;
    propValue_Current = propValue_CurrentTree;
    propValue_Read = propValue_ReadTree;

  }
```
</details>

al-fisher:
Thanks @alelom will review

alelom:
@al-fisher I think this is how I'd like to summarize the issue at point 2:

- The Explode is not behaving correctly in this case. This is summarized in https://github.com/BHoM/BHoM_UI/issues/241.
- There could be an argument that the above issue might be avoided by simplifying the `ModifiedPropertyPerObject` Dictionary to a specific oM type. However:
  - this does not remove the fact that there is something not right in the Explode.  
  - I'd like to avoid to add another "diff-type" object as it feels redundant in this case, but I would be happy to do it anyway if there is convergence. 
  - I think the Explode should be fixed regardless to cover cases like this.

PR #804, opened 2020-04-17
Reflection_oM: Provide a distinction between deleted and replaced instead of deprecated

### LINKED ISSUE #803: Reflection_oM: Provide a distinction between deleted and replace instead of deprecated
provide separated attributes to make the clear distinction of the two cases.
Deprecated would only be used as a generalisation of the two (i.e. `IsDeprecated()`)

@alelom , @al-fisher , @IsakNaslundBh , @FraserGreenroyd  FYI

adecler:

Closes #803

### DIFF
--- a/Reflection_oM/Attributes/ToBeDeletedAttribute.cs
+++ b/Reflection_oM/Attributes/ToBeDeletedAttribute.cs
@@ -30,9 +30,8 @@ using System.Threading.Tasks;
 
 namespace BH.oM.Reflection.Attributes
 {
-    [ToBeDeleted("3.2", "Use ReplacedAttribute or ToBeDeletedAttribute instead")]
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Enum)]
-    public class DeprecatedAttribute : Attribute, IImmutable
+    public class ToBeDeletedAttribute : Attribute, IImmutable
     {
         /***************************************************/
         /**** Properties                                ****/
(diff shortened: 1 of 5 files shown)


Reflection_oM/Attributes/ToBeDeletedAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Reflection.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Enum)]
+    public class ToBeDeletedAttribute : Attribute, IImmutable
```
FraserGreenroyd, line 34: Question, could this not just be `DeletedAttribute` short-handed to `[Deleted("3.2", "No one wanted this anymore")]` to align with the other attributes? Cause I could see an argument for the `Replaced` attribute to be a `HasBeenReplaced` if we're starting to put tenses into the names of attributes?

adecler: I was honestly going back and forth between the two so happy to follow the crowd. The idea behind using `ToBeDeleted` is that it is not deleted **yet** while `Replaced` has happened already.

FraserGreenroyd: That makes sense to me, might be worth hearing from @al-fisher and @IsakNaslundBh perhaps, but I'd say I'm happy to approve on Monday if there's no further discussion on the naming to get this merged in based on the above response :smile:

al-fisher: I think future tense is correct here. And is in fact the exception - so I personally have no problem with past tense for other attributes remaining implicit.
So would be happy with:

>  using ToBeDeleted is that it is not deleted **yet** while Replaced has happened already.

In terms of clarity of naming I wonder if `ToBeRemoved` contrasted with `Replaced` is clearer? 

Is subtle and potentially subjective - but I wonder if `TobeRemoved` might be clearer as referring to literally _removing_ functionality on deletion of a method - vs replacing functionality else where in the code.

adecler: Changed. Rolls better off the tong for some reason.

adecler:
@FraserGreenroyd , AppVeyor is happy but not DevOps. Not sure what is going on since everything is fine on the matching engine PR.

PR #805, opened 2020-04-20
Structure_OM -  Add Steel Connection object

### LINKED ISSUE #802: Structure_oM - Add Object for Steel Connection
#### Description:
Add an object for a structural element which represents connnections between steel framing objects

emidio-piermarini:
Closes #802 

Migration of Steel connections object from TeklaOM


Structure_oM/Elements/Bolt.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using BH.oM.Base;
+using BH.oM.Geometry;
+using System.ComponentModel;
+
+namespace BH.oM.Structure.Elements
+{
+    public class Bolt : BHoMObject
+    {
+        [Description("Represents the Diameter of the Bolt")]
```
rwemay, line 15: I would be more specific here as there are various 'diameters' - assume this is for the thread diameter?

rwemay, line 15: Add a description - assume this is for a generic bolt/any type (different heads, J-bolts, straight bolts, different thread lengths etc.).

Structure_oM/Elements/Bolt.cs
```diff
@@ -0,0 +1,23 @@
...
+        public virtual double Diameter { get; set; }
+
+        [Description("Represents Centerline of the Bolt from Top of Bolt Head to end of Shank")]
```
rwemay, line 18: If straight bolt then line is OK. If J-bolt we might just then need to add some parameters of the end of the bolt (instead of complicating this with ICurve).

Structure_oM/Elements/Cut.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Geometry;
+using System.ComponentModel;
+
+
+namespace BH.oM.Structure.Elements
+{
+    public class Cut : BHoMObject
```
rwemay, line 14: Add description, we probably want to describe what kind of cut this represents - assume this is intended to cut an element geometry with an extrusion perpendicular to the Curve geometry. The extrusion depth being equal to the thickness. (?) or something like that. We might want CutObject to be something more specific (IElement?) @IsakNaslundBh  ?

Structure_oM/Elements/SteelConnection.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Geometry;
+using BH.oM.Structure.Elements;
+using System.ComponentModel;
+
+namespace BH.oM.Structure.Elements
+{
+    public class SteelConnection : BHoMObject
```
rwemay, line 14: Add description. Presumably this is for a general steel connection (any number of elements) rather than (for example) a specific connection between two elements.

Structure_oM/Elements/SteelConnection.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        [Description("Framing Elements Connected Together by Connection")]
+        public virtual List<FramingElement> ConnectingElementIds { get; set; } = null;
```
rwemay, line 17: The name and type look like they're contradicting - is it a list of FramingElements or their Ids? If Ids are we using guids or some other identifier?

Structure_oM/Elements/Weld.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace BH.oM.Structure.Elements
+{
+    public class Weld : BHoMObject
+    {
+        [Description("Path of Root of Weld")]
+        public virtual ICurve WeldPath { get; set; }
+
+        [Description("Object to be Supported")]
+        public virtual BHoMObject ObjWelded { get; set; } = null;
```
rwemay, line 20: Similar to the cut, would we want to be specific here and use IElement, as BHoMObject obviously covers a huge number of objects.

Structure_oM/Elements/Weld.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+
+using BH.oM.Base;
+using BH.oM.Geometry;
+using System.ComponentModel;
+
+
+namespace BH.oM.Structure.Elements
+{
+    public class Weld : BHoMObject
```
rwemay, line 14: We probably want an IWeld at some point - for later. Add a description here - is this ALL welds - plug, spot, stitch, path or is it path based welds? If later, just be clear in description.

Structure_oM/Elements/SteelConnection.cs
IsakNaslundBh: can add to this, that the `FramingElement` (without the I) is deprecated, and replaced with the physical `IFramingElement`, so would not use that class at all.

al-fisher:
Just popping a `no-not-merge` on this PR until after the beta release, as classed as not a `severity:critical` `bug`

al-fisher:
PR lockdown for beta is now over - removing `do-not-merge`

IsakNaslundBh:
This has been stale for over 2 years now. All code in the PR exists on https://github.com/BHoM/TeklaStructures_Toolkit/tree/main/Tekla_oM/Elements

Will hence close this until more planning is in place for the classes to gain proper resource and review before being migrated.

PR #811, opened 2020-04-20
Spatial_oM: Adding in inital 2D-layout objects

### LINKED ISSUE #758: Spatial_oM: Add Layout objects
#### Description:

As discussed in https://github.com/BHoM/BHoM/issues/579

Need to add 2D-layout objects, controlling distributions of points in a region/along a curve. To be used for Reinforcement (as mentioned) but can also be used for things like pile caps etc.

2 main things to be ironed out before implementing is (and related).

1. Should these be BHoMObjects or not
2. Which oM should they belong to. Geometry/Spatial feels like the most intuitive, but could be convinced otherwise. @al-fisher @pawelbaran @adecler @rwemay @FraserGreenroyd 

Proposed objects/interfaces:

```
    [Description("Base interface for all 2D layouts")]
    public interface ILayout2D : IBHoMObject
    {

    }
```

```
    [Description("Base interface for all 2D layouts")]
    public class PerimiterLayout : BHoMObject, ILayout2D
    {

        [Description("Number of points along the perimiter of the host object")]
        public int NumberOfPoints { get; set; }

    }
```

```
    [Description("Layout for freeform control over the point distribution")]
    public class ExplicitLayout : BHoMObject, ILayout2D//, IImmutable //Maybe immutable, to make sure the points are always in xy-plane via create methods
    {

        [Description("The explicit grid of points in the layout")]
        public List<Point> Points { get; set; }

    }
```

```
    [Description("Linear distribution of points along a vector from one side of the perimiter of the host object to the other.")]
    public class LinearLayout : BHoMObject, ILayout2D//, IImmutable //Maybe immutable, to make sure the points are always in xy-plane via create methods
    {

        [Description("Number of points along the axis.")]
        public int NumberOfPoints { get; set; }

        [Description("Direction of the axis.")]
        public Vector Direction { get; set; }

        [Description("Offset of the linear layout in relation to the centroid of the host object.")]
        public Vector Offset { get; set; }

    }
```

IsakNaslundBh:

Closes #758 

Adding initial 2D-layout objects. Initial use is for reinforcement, but intended to be usefull across more cases/disciplines as well.

Will create followup Engine PR that makes use of the classes

### DIFF
--- a/Spatial_oM/Layouts/LinearLayout.cs
+++ b/Spatial_oM/Layouts/LinearLayout.cs
@@ -1,0 +1,63 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+
+using System.ComponentModel;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using BH.oM.Base;
+using BH.oM.Geometry;
+
+
+namespace BH.oM.Spatial.Layouts
+{
+    [Description("Linear distribution of points along a vector from one side of the perimiter of the host object to the other.")]
+    public class LinearLayout : BHoMObject, ILayout2D, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Number of points along the axis.")]
+        public int NumberOfPoints { get; }
+
+        [Description("Direction of the axis. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public Vector Direction { get; }
+
+        [Description("Offset of the linear layout in relation to the centroid of the host object. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public Vector Offset { get; }
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public LinearLayout(int numberOfPoints, Vector direction, Vector offset)
+        {
+            NumberOfPoints = numberOfPoints;
+            Direction = direction;
+            Offset = offset;
+        }
+
+        /***************************************************/
+    }
+}
--- a/Spatial_oM/Layouts/LinearLayout.cs
+++ b/Spatial_oM/Layouts/LinearLayout.cs
@@ -32,7 +32,7 @@ using BH.oM.Quantities.Attributes;
 
 namespace BH.oM.Spatial.Layouts
 {
-    [Description("Linear distribution of points along a vector from one side of the perimiter of the host object to the other.")]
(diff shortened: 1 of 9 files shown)


Spatial_oM/Layouts/LinearLayout.cs
```diff
@@ -0,0 +1,63 @@
...
+        /***************************************************/
+
+        [Description("Number of points along the axis.")]
+        public int NumberOfPoints { get; }
+
+        [Description("Direction of the axis. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public Vector Direction { get; }
+
+        [Description("Offset of the linear layout in relation to the centroid of the host object. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public Vector Offset { get; }
```
kThorsager, line 48: `Length` ?

Spatial_oM/Layouts/PerimiterLayout.cs
```diff
@@ -0,0 +1,41 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+
+using System.ComponentModel;
+using BH.oM.Base;
+
+namespace BH.oM.Spatial.Layouts
+{
+    [Description("Base interface for all 2D layouts")]
```
peterjamesnugent, line 29: Linear distribution of points along the perimiter of the host object.

Spatial_oM/Layouts/PerimiterLayout.cs
```diff
@@ -0,0 +1,41 @@
...
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Number of points along the perimiter of the host object")]
+        public int NumberOfPoints { get; set; }
+
+        /***************************************************/
+    }
+}
```
peterjamesnugent, line 41: Should we not have a constructor for a PerimeterLayout?

IsakNaslundBh: Not needed, as not immutable. Only added it to the other ones as I wanted to have create methods for them making sure the input geometry was valid.

al-fisher:
@IsakNaslundBh - for codeowners suggest same as Engine. 
@IsakNaslundBh @pawelbaran @al-fisher ?

IsakNaslundBh:
> `PerimEterLayout`
> 
> Besides that, I would think of adding to `PerimeterLayout` a perimeter domain or any other way of defining layout only on part of the perimeter?

Hmm, not a bad idea. Wonder if it is the same object, or another one?

Spatial_oM/Layouts/LinearLayout.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        [Description("Number of points along the axis.")]
+        public virtual int NumberOfPoints { get; set; }
+
+        [Description("Direction of the axis. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public virtual Vector Direction { get; }
+
+        [Length]
+        [Description("Offset of the linear layout in relation to the reference point, perpendicular to the Direction vector in the XY plane.")]
+        public virtual double Offset { get; }
```
kThorsager, line 50: Should specify what positive values vs negative values imply

Spatial_oM/Layouts/MultiLinearLayout.cs
```diff
@@ -0,0 +1,80 @@
...
+        [Length]
+        [Description("Minimum distance between any two layers.")]
+        public virtual double PerpendicularMinimumSpacing { get; set; }
+
+        [Description("Direction of the axis. Vector should lie in the XY-plane, i.e. have a Z-coordinate equal to 0.")]
+        public virtual Vector Direction { get; }
+
+        [Length]
+        [Description("Offset of the linear layout in relation to the reference point, perpendicular to the Direction vector in the XY plane.")]
+        public virtual double Offset { get; }
```
kThorsager, line 59: ^^

IsakNaslundBh:
> Not sure, gut feeling is same object. Not sure, however, what would be best way of describing the domain to make it practical. I think it might make sense to first see how does the current PR work in practice and see if adding yet another layer of complexity is actually necessary.
>
> So good to merge on my end 👍

Cool, will fix the spelling stuff, and agree, this already grew a lot more than I first expected, and can always be added later!

Spatial_oM/Layouts/LinearLayout.cs
IsakNaslundBh: Good one. Done.

IsakNaslundBh:
Had to update from master, as this is now lagging behind BHoM and BHoM_UI on master will not build without the changes.

al-fisher:
@FraserGreenroyd @pawelbaran @peterjamesnugent looks like your requested changes have been addressed
@IsakNaslundBh is that right? 
Can we confirm ready for merge - thanks 😄

al-fisher:
Thanks guys

PR #823, opened 2020-04-22
Reflection_oM: Adding PreviousVersion attribute for versioning support

adecler:
See https://github.com/BHoM/BHoM_UI/pull/247 for more details

Simply add a new attribute that will enable local versioning as an attribute directly applied to a method or property

### Test files
You will have to try you own here if you want to test. But you probably don't need to as it is just a very simple object definition.

### DIFF
--- a/Reflection_oM/Attributes/OldVersionAttribute.cs
+++ b/Reflection_oM/Attributes/OldVersionAttribute.cs
@@ -1,0 +1,60 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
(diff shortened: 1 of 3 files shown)


adecler:
Yes, we should add constructor to the list of authorised targets. I also think that the attribute should be named 'PreviousVersion' instead.

Reflection_oM/Attributes/OldVersionAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Reflection.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property)]
+    public class OldVersionAttribute : Attribute, IImmutable
```
FraserGreenroyd, line 34: To be `PreviousVersion` as discussed

Reflection_oM/Attributes/OldVersionAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Reflection.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property)]
```
FraserGreenroyd, line 33: To include `AttributeTargets.Constructor` as discussed

adecler:
@FraserGreenroyd , I have made the changes discussed. I have also added `AllowMultiple = true` so we can keep all the previous version if necessary. My initial idea was to keep only the ones for the current version (so batch remove at the start of each quarter) but this might be a bit impractical and, even then, there is not guarantee that two methods would not be replaced by the same one.

Reflection_oM/Attributes/OldVersionAttribute.cs
adecler: done

PR #839, opened 2020-04-30
Graphics_oM: additions for the new ToRenderMesh infrastructure

### LINKED ISSUE #838: Graphics_oM: addition for new ToRenderMesh infrastructure
#### Description:

Changes needed for https://github.com/BHoM/BHoM_Engine/issues/1694

alelom:

Closes #838 

### Test files

See the test file in https://github.com/BHoM/TriangleNet_Toolkit/pull/27 to see how it works.
Also see test file in https://github.com/BHoM/Speckle_Toolkit/pull/86


Graphics_oM/Render/RepresentationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+        public virtual bool Cap1DElements { get; set; } = false;
+
+        [Description("If true, 2D elements are detailed. E.g. panels are represented as boxes with their thickness.")]
+        public virtual bool Detailed2DElements { get; set; } = false;
+
+        [Description("The key of the BHoMObjects' CustomData dictionary where a custom RenderMesh might be found.")]
+        public string CustomRendermeshKey { get; set; } = "Rendermesh";
+
+        [Description("The key of the BHoMObjects' CustomData dictionary where colour information might be found.")]
+        public string CustomDataColorKey { get; set; } = "Colour";
```
FraserGreenroyd, line 60: These properties need the `virtual` keyword

Graphics_oM/Render/RepresentationOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    [Description("Defines options for the Representation of BHoM Objects.")]
+    public class RepresentationOptions : BHoMObject
+    {
+        /***************************************************/
+        /****            Public Properties              ****/
+        /***************************************************/
+
+        [Description("If true, 0D elements are detailed. E.g. Points are represented by spheres; Node representation includes the support condition (e.g. pyramid with sphere on top for a Pin).")]
+        public virtual bool Detailed0DElements { get; set; } = true;
```
rwemay, line 39: Assume you've done a lot more thinking on this than me recently(!) but wondering if boolean is too limiting for future use. Might it be better to follow the 'level of detail' approach quite heavily adopted in BIM? Could be DetailLevel0D, DetailLevel1D etc, and then an enum or integer value to begin with?

Graphics_oM/Render/RepresentationOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Description("Scale the Element0D representation.")]
+        public virtual double Element0DScale { get; set; } = 1;
+
+        [Description("If true, 1D elements are detailed: e.g. the Bars are extruded using their Section property. Else, the Bars' centreline is coarsely piped.")]
+        public virtual bool Detailed1DElements { get; set; } = true;
```
rwemay, line 45: Given the scale of what we're going to use detail level for, I'd be tempted to remove reference to specific objects within disciplines or analytical definitions, and rather refer to a wiki on level of detail.

Graphics_oM/Render/RepresentationOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        [Description("Scale the Element1D representation, if applicable. E.g. for Lines, this increases the Pipe size.")]
+        public virtual double Element1DScale { get; set; } = 1;
+
+        [Description("Cap 1D elements, if applicable. E.g. for a Line, this caps the computed Pipe; for a Bar it may cap the Extrusion of its cross section.")]
+        public virtual bool Cap1DElements { get; set; } = false;
```
rwemay, line 51: Was wondering if this needed to be limited to 1D rather than just 'cap elements' - as I guess we could cap other meshes with open edges in the future (so could just be `CapElements`. Not sure. In any case, I'd remove reference to `Bar` and/or make it discipline agnostic - cap the extrusion might suffice.

Graphics_oM/Render/RenderMeshOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /***************************************************/
+        /****            Public Properties              ****/
+        /***************************************************/
+
+        [Description("Regulate how the Representation of the objects is computed.")]
+        public virtual RepresentationOptions RepresentationOptions { get; set; } = new RepresentationOptions();
+
+        [Description("Regulates the refinement of Pipe meshes for 1DElements, i.e. the number of faces of the Pipes.")]
+        public virtual double Element1DRefinement { get; set; } = 1;
```
rwemay, line 42: If this is number of faces, should be integer(?)

Graphics_oM/Render/RenderMeshOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [Description("Regulates the refinement of surface meshes for 2DElements.")]
+        public virtual double Element2DRefinement { get; set; } = 1;
+
+        [Description("The key of the BHoMObjects' CustomData dictionary where a custom RenderMesh might be found.")]
+        public string CustomRendermeshKey { get; set; } = "Rendermesh";
```
rwemay, line 48: Should be `CustomRenderMeshKey` and `RenderMesh`?

Graphics_oM/Render/RepresentationOptions.cs
alelom: @rwemay thanks for your comments, more than happy to have your feedback!  😄 

I get all of your points, happy to discuss it later during our call – I think we will be able to agree on something much faster.

Cheers!

Graphics_oM/Render/RepresentationOptions.cs
alelom: Agreed, will do this now!

Graphics_oM/Render/RenderMeshOptions.cs
alelom: I updated the description to explain the behaviour. Have a look at the test file in https://github.com/BHoM/TriangleNet_Toolkit/pull/27 to see how it works.

Graphics_oM/Render/RepresentationOptions.cs
alelom: Summarising my thoughts - yes I agree but as a prototype I believe this suffices for now. To add that kind of flexibility, we will see later as uses arise. 

Please have a look at the test file in https://github.com/BHoM/TriangleNet_Toolkit/pull/27 to see how it works.

Issue #850, opened 2020-05-04
Geometry_oM: Add new wiki page to document the explicit casts between geometry primitives


al-fisher:
#### What is missing/incorrect?

As discussed on Foundational Sprint Kick-Off call just now - as we plan to implement some more critical explicit casts - good to create clear documentation of when and why they are implemented.

Agreed a nice matrix showing for each pairing of primitives - highlighting where explicit casts do exist - and with a notes column explaining logic of primary reason it is needed or indeed not implemented/excluded.

@IsakNaslundBh @adecler FYI

kThorsager:
Made a first draft on a wiki page:
https://github.com/BHoM/BHoM/wiki/Geometry_oM:-Explicit-casts
Think it became quite good, but please do check it out and complain wherever necessary

al-fisher:
Very useful and comprehensive. Thanks @kThorsager - happy to close this as certainly addressed.

Issue #789, opened 2020-04-14
Geometry_oM: Add explicit casts between curve types.


IsakNaslundBh:
#### Description:

To help the UI out for some curve types, that could be defined in many different ways with the same result, suggesting to add a few explicit casts in the Geometry_oM between the BHoM curves.

for example

- Line -> PolyLine
- Line -> NurbsCurve
- Line -> PolyCurve
- PolyLine -> PolyCurve
- PolyLine -> NurbsCurve

and potentially the other way around (with checks).

This would make the coverts in the UIs a lot simpler to deal with, as you only have to return one thing (maybe the most simple possible) and can let the explicit cast to what the method wants.

Would hopefully help solve issues like 2-noded `Polylines` not feedable to a BHoMComponent expecting a `Line` etc.

@adecler can you see any downside with adding this?

@pawelbaran @al-fisher @adecler @FraserGreenroyd @epignatelli @alelom @kThorsager

FraserGreenroyd:
Title says Geometry_Engine but this is posted on BHoM so I'm guessing that's just a typo?

I'm not convinced this is the right solution in terms of the BHoM philosophy, we've managed to survive this long without these converts, why do we need to add them now?

Why can't they be convert methods within the Geometry_Engine which then makes it very explicit to the user what they can and cannot convert between. This also prevents any 'accidental' converts happening - the user must be very conscious of dropping a convert component on a canvas.

That's my tuppence anyway, but I'm not a big geometry maintainer so happy to leave the big guns to the debate mostly!

pawelbaran:
> I'm not convinced this is the right solution in terms of the BHoM philosophy, we've managed to survive this long without these converts, why do we need to add them now?

Because life should not be about survival, but about comfort (even though it might look a bit different recently).

> Why can't they be convert methods within the Geometry_Engine which then makes it very explicit to the user what they can and cannot convert between. This also prevents any 'accidental' converts happening - the user must be very conscious of dropping a convert component on a canvas.

Well, the need for explicit convert is the actual issue here @FraserGreenroyd - the end user does not want to bother about the difference between 2-point nurbs curve, 2-point polyline and a line. Plus, Grasshopper does already so many 'accidental' converts that it will not make too much difference if we add a few 😆

adecler:
@IsakNaslundBh , I have no problem with doing that.

[...]

PR #851, opened 2020-05-05
Geometry_oM: Explicit casts for Curves

### LINKED ISSUE #790: Geometry_oM: Polyline casts
#### Description:

As part of https://github.com/BHoM/BHoM/issues/789 and in support of https://github.com/BHoM/Dynamo_Toolkit/issues/224

Suggesting adding explicit casting:

- Polyline -> PolyCurve
- Polyline -> NurbsCurve

kThorsager:
Closes #790 
Think I'll leave #789 open as it also details step two, the casts with checks
related to #850 which has the wiki page

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EQBbBPE6BYFBlhJwbJllVxIB_ADwYP4S2qo4S03W9DWiwA?e=4BpRMQ

### Changelog

- `ICurve` to `PolyCurve`
- `Line`, `Polyline`, `Ellipse`, `Circle` to `NurbsCurve`
- `Circle` to `Ellipse`
- `Line` to `Polyline`

### Additional comments

There were mentions about various UI things in the issue #789 which I think was decided as not a problem. But could be worth double checking.

### DIFF
--- a/Geometry_oM/Curve/PolyCurve.cs
+++ b/Geometry_oM/Curve/PolyCurve.cs
@@ -34,7 +34,51 @@ namespace BH.oM.Geometry
 
         [Description("A collection of curves, of any or mixed type, which together define the composite shape.")]
         public virtual List<ICurve> Curves { get; set; } = new List<ICurve>();
-        
+
+        /***************************************************/
+        /**** Explicit Casting                          ****/
+        /***************************************************/
+
+        public static explicit operator PolyCurve(Arc curve)
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
+
+        /***************************************************/
+
+        public static explicit operator PolyCurve(Circle curve)
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
+
+        /***************************************************/
+
+        public static explicit operator PolyCurve(Ellipse curve)
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
+
+        /***************************************************/
+
+        public static explicit operator PolyCurve(Line curve)
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
+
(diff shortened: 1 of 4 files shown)


Geometry_oM/Curve/PolyCurve.cs
```diff
@@ -34,7 +34,51 @@ public class PolyCurve : ICurve
...
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
+
+        /***************************************************/
+
+        public static explicit operator PolyCurve(Polyline curve)
+        {
+            return new PolyCurve() { Curves = new List<ICurve>() { curve } };
+        }
```
pawelbaran, line 80: This should query `Polyline` for its segments first?

kThorsager:
Yes, my plan was to implement all the unconditional casts first and follow up with those requiring checks.
Could do those now as well if you'd like?

Geometry_oM/Curve/PolyCurve.cs
kThorsager: A mix of copy paste and lack of query methods in the oM, but can make it happen

pawelbaran:
No opinion on the sequence, as long as we get this sorted :) happy to approve once my change request is resolved.

Geometry_oM/Curve/PolyCurve.cs
pawelbaran: Yes please - I think the current implementation is going to do more harm than good:
image

:rofl:

pawelbaran:
LGTM now, would be nice to get the missing ones sorted as soon as this is merged (to make sure this does not fall under the radar)

al-fisher:
> Some are ofc a bit more comlex than one would hope to ahve in the oM, but given the functionality they provide I think well worth it, and it looks like you have done a good job of keeping the code to the minimum.

Agreed. 
Good job @kThorsager 👍

Issue #843, opened 2020-05-01
Data_oM: Spatial data class for data with extents


kThorsager:
#### Description:
Will start of by saying that I'm quite biased towards creating some kind of `RTree`

I know of `PointMatrix`

- Mainly dislike that it is very dependent on the initial cell size.
- There's also the difficulty in adapting it for data with extent (i.e. `ICurves`) as it is defined for Points. Now, I do think that it's possible, but it strikes me as building out on mismatched foundation.
- No "good" solution for closest item, as the solution without a max distance would involve stepping out cell by cell until it found something, which would be very inefficient if the closest item isn't close to the search point. 
(And the uninitiated might take it as more of an arbitrary upper bound and set it as distance from the search point to the first point to search among. Both causing issues when it tries to sort every point in the search set and when it tries to Query every empty cell between the search point and the points to search among if one say happened to pull a Revit model in the wrong coordinate system and searched from a static source)

A spatial data tree would also have some initial values to consider, namely:

- Sub branches per branch
- number of items in the end branches

But they won't be as scale dependant and a faulty value won't be as detrimental as for the `PointMatrix`. That is as the worst case would be having every item in one branch, which would equate to not having an tree structure.
A faulty value for the `PointMatrix` could lead to the need to check an arbitrary amount of cells. 
(which isn't really a fair comparison as that would be an extreme case which most in-code users would safeguard from, but difficult to keep from being able to happen centrally)

So would propose adding an Spatial Data Tree class which can handle elements with extents and would be happy to implement it.

Otherwise I'll have a look at fixing up the `PointMatrix` situation.

@pawelbaran @al-fisher @adecler Give me a go ahead if this is fine, and add anyone else who gets a say in this

adecler:
I have no issue with you adding an alternative to the `PointMatrix`. As always, it is great to have multiple options.

Adding @rolyhudson as he might be interested too.

al-fisher:
Yes - agreed! Sounds good to me too

rolyhudson:
This sounds great. I can see this being really useful for geospatial scale queries too.

PR #852, opened 2020-05-05
Data_oM: Prototype Data Tree for data with extents with create and query methods based on multiple numerical domains

kThorsager:
## Prototype
This is a prototype and should not be considered in regards to blocking, as these branches won't be merged. This PR is opened to share and discuss code.

 #843 
Adds the base objects for a `NTree` data structure which is based on a `RTree` concept and generalised for an arbitrary amount of dimensions.

The use of generics here has been by trial and error, please notify if improvements can be made.


Data_oM/Collections/NBound.cs
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+using System.Collections.Generic;
+using BH.oM.Base;
+using System.Linq;
+using System;
+
+namespace BH.oM.Data.Collections
+{
+    public class NBound : IObject
```
kThorsager, line 30: I added a `NBound` object to make this very generic and avoid dependencies on Geometry_oM.

So first of all, the name. It's a `BoundingBox` for an arbitrary amount of dimensions N. Does it work?

And does it go in `Data.Collections`? I saw that `LocalData` did which also is a support object for a collection but feels slightly strange.

Data_oM/Collections/NBound.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public double[] Min { get; set; } = null;
+
+        public double[] Max { get; set; } = null;
```
kThorsager, line 34: Does it need to be `IImmutable` as one could set separate length arrays?

One could also set min values which are higher than corresponding max values, but that's a problem for `BoundingBox` as well.

Data_oM/Collections/NTree.cs
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System.Collections.Generic;
+using BH.oM.Base;
+
+namespace BH.oM.Data.Collections
+{
+    public class NTree<T> : ITree
```
kThorsager, line 28: The name here as well, its a data tree for `NBounds`, `NTree`. Also not that far of from say `RTree` which is what it's based on.

Data_oM/Collections/NTree.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public NBound Bounds { get; set; } = null;
+
+        public ITree[] Items { get; set; } = null;
+    }
```
kThorsager, line 33: `NTree` does currently not have a "root" object which keeps track of things like the size of each level or the size of the leaf level.

Thought that I did not really need it apart from modifying the collection of an existing `NTree` and that it was more in line with other object in the BHoM.

Think one can Query an approximate size from an existing `NTree` so hopefully not an big issue.

Data_oM/Collections/NTree.cs
kThorsager: There's also the perhaps silly thought of making it a `IEnumerable` since it is a collection.

Think that would allow one to check if a incoming collection is a `NTree` in the Create methods for it and hence cut out the build time from processes which use it on multiple levels.

But I've not played around that much with that and don't really know what drawbacks there would be to that approach.

kThorsager:
@pawelbaran @al-fisher @rolyhudson @adecler 

Would still need some comments on my comments about general approach.
Will remind that a full review is not needed at this stage.

Otherwise this is done apart from a bunch of descriptions (and potentially refinement of method names)

Data_oM/Collections/NBound.cs
pawelbaran: I would go for `IImmutable` - the earlier we do that the less changes later?

pawelbaran:
Sorry if I am derailing the idea (was not involved from the beginning), but I would rather keep `ITree` free of any properties. `Bounds` already imposes that we are talking of a spatial tree - what if we want to build a simple binary tree?

Or am I missing a point?

kThorsager:
> Sorry if I am derailing the idea (was not involved from the beginning), but I would rather keep `ITree` free of any properties. `Bounds` already imposes that we are talking of a spatial tree - what if we want to build a simple binary tree?
> 
> Or am I missing a point?

Having Bounds on `ITree` helps a lot in the code, as it means that every `NTree` can be defined as something which encompasses a collection of other extents.

But I do see your point about `ITree` seeming like an interface and hence "occupy" that name whenever someone else implements some kind of tree.

So I would still want something which is an interface for the two components in the `NTree`, but guess it is a object specific interface kind of, now that I think about it.

What is the naming convention for those kind of interfaces?

rolyhudson:
Quite excited by the possibilities here especially the n-dimensional potential when extending to broader data analysis beyond 3d model spaces.

The N in the names is confusing me 🙃 a little bit !
 - `NBound` is an n-dimensional bounding box like object
 - `NTree` is a tree where each node has N children (n-ary tree?))

Should the tree not just be defined as Nodes structured with Generics like: 
```
public class Node<T>
{
    public virtual T Value { get; set; } = default(T);
    public List<Node<T>> Children {get; set;}
}
```
`Node<NBound> root = new Node<NBound>()` would set up the `NTree`

And we would also get a generic n(m)-ary tree for others to use

`NBound` is a store of multiple max - mins that works for any set of n-properties.

Would it be easier to have a `Range` object for each property?

Otherwise it depends on ordered arrays of Max and Min
So maybe the `NBound` has the property is `List<Range> Ranges`?

pawelbaran:
> But I do see your point about `ITree` seeming like an interface and hence "occupy" that name whenever someone else implements some kind of tree.

Maybe we could have `ISpatialTree : ITree`?

kThorsager:
> Quite excited by the possibilities here especially the n-dimensional potential when extending to broader data analysis beyond 3d model spaces.

😃

> The N in the names is confusing me 🙃 a little bit !
> 
> * `NBound` is an n-dimensional bounding box like object
> * `NTree` is a tree where each node has N children (n-ary tree?))

I was thinking NTree as containing things in N-dimensions thinking more from the name RTree, but now see the confusion as things like binary-trees refer to number of branches.

> Should the tree not just be defined as Nodes structured with Generics like:
> 
> ```
> public class Node<T>
> {
>     public virtual T Value { get; set; } = default(T);
>     public List<Node<T>> Children {get; set;}
> }
> ```

That looks like you'd get a value at each "level" of the Tree, where I've gone for keeping all the data at the "bottom" of the Tree.
But maybe leaving that empty or using it wouldn't be a problem

> `Node<NBound> root = new Node<NBound>()` would set up the `NTree`

Specifically like that looks annoying as one wouldn't have anywhere to keep the data, but inheriting `Node` and adding a `NBound` property should be able to do something similar.

> And we would also get a generic n(m)-ary tree for others to use

These Tree prefixes are less clear than I thought, I'm not sure what you're meaning here

> `NBound` is a store of multiple max - mins that works for any set of n-properties.
> 
> Would it be easier to have a `Range` object for each property?
> 
> Otherwise it depends on ordered arrays of Max and Min
> So maybe the `NBound` has the property is `List<Range> Ranges`?

This does sound a lot better, 
[and had to check to make sure on this](https://www.varsitytutors.com/hotmath/hotmath_help/topics/domain-and-range), but think we'd want to call it `Domain` to avoid confusion. (But may not be so clear cut as a `BoundingBox` hardly is a function)

rolyhudson:
> That looks like you'd get a value at each "level" of the Tree, where I've gone for keeping all the data at the "bottom" of the Tree.
But maybe leaving that empty or using it wouldn't be a problem

I was thinking the `Value` property is the NBounds as in the rectangles or boxes in an RTree.  The data is stored on leaf nodes (potentially as a reference only) right?

> Specifically like that looks annoying as one wouldn't have anywhere to keep the data, but inheriting Node and adding a NBound property should be able to do something similar.

Maybe there should be an `INode` which is implemented by a general `Node` (root and mid tree) and `Leaf` that  adds a data reference property.

> These Tree prefixes are less clear than I thought, I'm not sure what you're meaning here

Just that if you are going to set up a data structure might as well make it general as I'm sure other will have a future use.

Not sure https://www.varsitytutors.com/hotmath 🔥 is a solid reference, but I agree the terminology and usage of `Domain` and `Range` should be correct.

kThorsager:
> > That looks like you'd get a value at each "level" of the Tree, where I've gone for keeping all the data at the "bottom" of the Tree.
> > But maybe leaving that empty or using it wouldn't be a problem
> 
> I was thinking the `Value` property is the NBounds as in the rectangles or boxes in an RTree. The data is stored on leaf nodes (potentially as a reference only) right?
> 
> > Specifically like that looks annoying as one wouldn't have anywhere to keep the data, but inheriting Node and adding a NBound property should be able to do something similar.
> 
> Maybe there should be an `INode` which is implemented by a general `Node` (root and mid tree) and `Leaf` that adds a data reference property.

Ok, I think I see. Think I'm just a bit against in that case keeping an empty list of sub Nodes on the leaves

> Not sure https://www.varsitytutors.com/hotmath 🔥 is a solid reference, but I agree the terminology and usage of `Domain` and `Range` should be correct.

Top result on Google, not sure how much more of a solid reference you can get 😜

kThorsager:
So maybe to resolve naming:
new object `Domain`, (containing a min and max value)
`NBound` => `DomainBox`, (fairly sure box is used in an arbitrary amount of dimensions),
`NTree` => `DomainTree`, as it's a Tree based on Domains

Still thinking on how to restructure the (as of now) `ITree` objects

rolyhudson:
A thought on the name of this PR: Use of _Spatial_ suggests limited dimensions (3?) for something that is created as n-dimensional.

Also thinking about the name options `NTree` or `DomainTree`. I would consider implementing the generic n-ary tree as discussed previously. Not binding to the `DomainBox` type would allow other use cases.

kThorsager:
> Maybe there should be an INode which is implemented by a general Node (root and mid tree) and Leaf that adds a data reference property.

Have done some frantic googling and wikipedia-ing on Trees outside of RTrees (and similar).
So I interpret that as this?
``` C#
public interface INode<T> : IDataStructure
{
    List<INode<T>> Children { get; set; }

    T NodeValue { get; set; }
}

public class Node<T> : INode<T>
{
    public virtual List<INode<T>> Children { get; set; }

    public virtual T NodeValue { get; set; }
}

public class LeafNode<T, F> : INode<T>
{
    public virtual List<INode<T>> Children { get; set; }

    public virtual T NodeValue { get; set; }

    public virtual F DataValue { get; set; } 
}
```
Fist minor thing but `INode` seems redundant.

But my bigger issue would be that you wouldn't have the `DataValue`s type `F` available from the root. So I would know that it is based on `NBound`s but not if it contains `Bar`s or `Mesh`es and can hence not return items of that type for queries.
Adding The extra argument higher up without the property seems silly as one would have to define a type which isn't used for every other implementation.

> Also thinking about the name options NTree or DomainTree. I would consider implementing the generic n-ary tree as discussed previously. Not binding to the DomainBox type would allow other use cases.

Yes, working on some good solution.
But was saying that as I was thinking if one does this kind of inheritance.
``` C#
public class Node<T> : IDataStructure
{
    public virtual List<Node<T>> Children { get; set; }

    public virtual T NodeValue { get; set; }
}

public class DomainNode<T> : Node<T>
{
    public virtual DomainBox Domain { get; set; }
}
```

And there's also the option of flipping the steak, and have one kind where the children are related to some data (NBound/DomainBox in this case), and that the other kind is inheriting like this.
``` C#
public class Node<F,T> : IDataStructure
{
    public virtual List<Tuple<F, Node<F,T>>> Children { get; set; }

    public virtual T NodeValue { get; set; }
}

public class Node<T> : Node<string, T>
{
    public new List<Tuple<string, Node<string, T>>> Children { get; set; }
}
```

These are very untested ideas I'm not so sure of myself yet.

rolyhudson:
> Fist minor thing but INode seems redundant.

Yes but remove the property `List<INode<T>> Children { get; set; }` and your leaf does not need to implement it which I thought was what you wanted to avoid?

Although not having any children is quite a useful way of knowing you are on a leaf when traversing. 

Seems like its getting too complicated though.

Your last proposal above made me look again at the `Tree` that is already in the BHoM.

What about adding an attributes property to the existing `Tree` and using that for your data at the leaves? We recently added attributes to `GraphNode` could be useful here too?

kThorsager:
> > Fist minor thing but INode seems redundant.
> 
> Yes but remove the property `List<INode<T>> Children { get; set; }` and your leaf does not need to implement it which I thought was what you wanted to avoid?
 >
> Although not having any children is quite a useful way of knowing you are on a leaf when traversing.

I'm warming up to the idea, especially after doing a second less frantic tour of Wikipedia

> Seems like its getting too complicated though.
> 
> Your last proposal above made me look again at the `Tree` that is already in the BHoM.
> 
> What about adding an attributes property to the existing `Tree` and using that for your data at the leaves? We recently added attributes to `GraphNode` could be useful here too?

A full Dictionary seems like a little overkill for storing a data value.
But I haven't looked to much on the existing Tree, and am a bit worried about messing with it as I'm realising the width of the definition. But can see that @epignatelli has made a implementation's of it in DeepLearning_oM. Do you have things to add to this discussion? (sorry for bringing you in this late)

But think I have a new suggestion.

Add a base class which is the definition of a Tree, i.e. A value with children defined recursively.
``` C#
public class Node<T> : IDataStructure
{
    public virtual IEnumerable<Node<T>> Children { get; set; } = new List<Node<T>>();

    public virtual T Value { get; set; } = default(T);
}
```
Add a second class with a Relation value which isn't necessarily data but more defining of the Tree's edges property (ish, WIP)
``` C#
public class Node<R, T> : Node<T>
{
    public new IEnumerable<Node<R, T>> Children { get; set; } = new List<Node<R, T>>();

    // Not needed, but to clarify that it's still around
    public override T Value { get; set; } = default(T);

    [Description("The relation between this Node and either its parent, children and/or sibling.")]
    public virtual R Relation { get; set; } = default(R);
}
```
Can also add specific implementations in classes to alleviate UI/writing the code
``` C#
public class DomainTree<T> : Node<DomainBox, T>
{ }
```
Think it might be beneficial to at least have the ability to define a root class with the definition of the Tree's construction.
``` C#
public class Tree<R,T> : Node<R, T>
{
    public virtual TreeConfig Config { get; set; } = new TreeConfig();
}
```
This is the Current Tree.
``` C#
public class Tree<T> : IDataStructure
{

    public virtual Dictionary<string, Tree<T>> Children { get; set; } = new Dictionary<string, Tree<T>>();

    public virtual string Name { get; set; } = "";

    public virtual T Value { get; set; } = default(T);


}
```
And I'm fairly sure that it could be redefined a one of these with the new base class. Depending on what the dictionary `string` is for and ability to set the comparer for the `HashSet` to the specific string.
``` C#
public class Tree<T> : Node<string, Tuple<string, T>>
{
    public new HashSet<Tree<T>> Children { get; set; } = new HashSet<Tree<T>>();
}

public class Tree<T> : Node<Tuple<string, string, T>>
{
    public new HashSet<Tree<T>> Children { get; set; } = new HashSet<Tree<T>>();
}
```

> Seems like its getting too complicated though.

Not sure this helped much, but think it should be fairly straight forward when one gets into it (hope).

But hope I made sense at least

kThorsager:
I think I have refactored this while keeping the key bits.
- The collection should be able to return the data in their original type
- Simple to use, (Using the DomainTree, not writing methods for it)
- More generalised to facilitate other similar development
- (still works)

Key changes:
- `NBound` -> `DomainBox`, which has an array of `Domains`
- Added `INode<T>` as a interface for tree like data structures and has it relate to the actual `DomainTree<T>` in a similar manner to how `IElement` works. There are two methods in the engine which are needed for the interface methods to work. `Values()` return the current nodes values, and `Children` return the current nodes children.
- `DomainTree<T>` is now a single object with both values and children on each node.

Think this will keep it fairly modular while also not making it completely confusing.

And then a note to @adecler . I did kind of break the UI again when making `DomainTree<T>` an `IEnumereble<T>`. Now this is hardly a critical part to make this work but more of a nice thing to add. And while I couldn't find much precedent for it, I also couldn't find anything telling me I couldn't.
Is it allowed at all?

But given that these changes get some approval I'll get to adding descriptions, (and presumably opening a fresh branch for a smoother merge) and prepare it for a pre-merge review.

kThorsager:
Closing this as #914 is the final version

PR #855, opened 2020-05-07
Geometry_oM: Add Conditional Explicit Casts Between Curve Types.

### LINKED ISSUE #789: Geometry_oM: Add explicit casts between curve types.

kThorsager:

Closes #789 
There's also the wiki created in #850 

Adds all the explicit casts for curves which has a failure state. 

### Test files

The right hand part of the test, which is testing that nothing breaks to bad (i.e. throws an exception), does not work as the UI does that without https://github.com/BHoM/BHoM_UI/issues/255 s fix, which is handled in https://github.com/BHoM/BHoM_UI/pull/253
https://burohappold.sharepoint.com/:u:/s/BHoM/EUt5Ob8QoL9BtmZo-qw3QvABpAdElVF6QxXzRAR8BmY9lA?e=d6pNRi

### Changelog

- from `PolyCurve` to any `ICurve`
- from `Ellipse` to `Arc`, `Circle`
- from `Arc` to `Circle`, `Ellipse`
- from `Polyline` to `Line`
- from `NurbsCurve` to `Line`, `Polyline`

### Additional comments

Worth mentioning that the reason it got a bit messy in casting from `PolyCurve` to something is that C# does not support user defined casting from Interfaces. which means that these casts will not be that easily used in the code as one might hope.

So one will only be able to cast specifically from a specific type to another specific type. Not from a interface to a type and expect a `null` if it failed.

### DIFF
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -95,22 +95,19 @@ namespace BH.oM.Geometry
                 return null;
 
             ICurve c = curve.Curves[0];
-            switch (c.GetType().Name)
-            {
-                case "Arc":
-                    return c as Arc;
-                case "Circle":
-                    Circle circle = c as Circle;
-                    return (Arc)circle;
-                case "Ellipse":
-                    Ellipse ellipse = c as Ellipse;
-                    return (Arc)ellipse;
-                case "PolyCurve":
-                    PolyCurve polyCurve = c as PolyCurve;
-                    return (Arc)polyCurve;
-                default:
-                    return null;
-            }
+            if (c == null)
+                return null;
+            else if (c is Arc)
+                return c as Arc;
+            else if (c is Circle)
+                return (Arc)(c as Circle);
+            else if (c is Ellipse)
+                return (Arc)(c as Ellipse);
+            else if (c is PolyCurve)
+                return (Arc)(c as PolyCurve);
+            else
+                return null;
+
         }
 
         /***************************************************/
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -95,9 +95,7 @@ namespace BH.oM.Geometry
                 return null;
 
             ICurve c = curve.Curves[0];
-            if (c == null)
-                return null;
-            else if (c is Arc)
+            if (c is Arc)
                 return c as Arc;
             else if (c is Circle)
                 return (Arc)(c as Circle);
--- a/Geometry_oM/Curve/Arc.cs
+++ b/Geometry_oM/Curve/Arc.cs
@@ -48,6 +48,7 @@ namespace BH.oM.Geometry
         [Description("Angle in radians to the end point from the local x-axis, counter clockwise around the local z-axis.")]
         public virtual double EndAngle { get; set; } = 0;
 
+
         /***************************************************/
         /**** Explicit Casting - Special Case           ****/
         /***************************************************/
@@ -84,6 +85,44 @@ namespace BH.oM.Geometry
                 StartAngle = 0,
                 EndAngle = Math.PI * 2
             };
+        }
+
+        /***************************************************/
+
+        public static explicit operator Arc(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
+            {
+                case "Arc":
+                    return c as Arc;
+                case "Circle":
+                    Circle circle = c as Circle;
+                    return (Arc)circle;
+                case "Ellipse":
+                    Ellipse ellipse = c as Ellipse;
+                    return (Arc)ellipse;
+                case "PolyCurve":
+                    PolyCurve polyCurve = c as PolyCurve;
+                    return (Arc)polyCurve;
+                default:
+                    return null;
+            }
+        }
+
+        /***************************************************/
+
+        public static explicit operator Arc(Ellipse curve)
+        {
+            Circle circle = (Circle)curve;
+
+            if (circle == null)
+                return null;
+
+            return (Arc)circle;
         }
 
         /***************************************************/
--- a/Geometry_oM/Curve/Circle.cs
+++ b/Geometry_oM/Curve/Circle.cs
@@ -53,22 +53,18 @@ namespace BH.oM.Geometry
                 return null;
             
             ICurve c = curve.Curves[0];
-            switch (c.GetType().Name)
-            {
-                case "Circle":
-                    return c as Circle;
-                case "Arc":
-                    Arc arc = c as Arc;
-                    return (Circle)arc;
-                case "Ellipse":
-                    Ellipse ellipse = c as Ellipse;
-                    return (Circle)ellipse;
-                case "PolyCurve":
-                    PolyCurve polyCurve = c as PolyCurve;
-                    return (Circle)polyCurve;
-                default:
-                    return null;
-            }
+            if (c == null)
+                return null;
+            else if (c is Circle)
+                return c as Circle;
+            else if (c is Arc)
+                return (Circle)(c as Arc);
+            else if (c is Ellipse)
+                return ((Circle)(c as Ellipse));
+            else if (c is PolyCurve)
+                return (Circle)(c as PolyCurve);
+            else
+                return null;
         }
 
         /***************************************************/
--- a/Geometry_oM/Curve/Circle.cs
+++ b/Geometry_oM/Curve/Circle.cs
@@ -42,6 +42,70 @@ namespace BH.oM.Geometry
         [Description("Distance from the Centre to any point on Circle.")]
         public virtual double Radius { get; set; } = 0;
 
+
+        /***************************************************/
+        /**** Explicit Casting - Special Case           ****/
+        /***************************************************/
+
+        public static explicit operator Circle(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+            
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
+            {
+                case "Circle":
+                    return c as Circle;
+                case "Arc":
+                    Arc arc = c as Arc;
+                    return (Circle)arc;
+                case "Ellipse":
+                    Ellipse ellipse = c as Ellipse;
+                    return (Circle)ellipse;
+                case "PolyCurve":
+                    PolyCurve polyCurve = c as PolyCurve;
+                    return (Circle)polyCurve;
+                default:
+                    return null;
+            }
+        }
+
+        /***************************************************/
+
+        public static explicit operator Circle(Ellipse curve)
+        {
+            if (System.Math.Abs(curve.Radius1 - curve.Radius2) > Tolerance.Distance)
+                return null;
+
+            Vector x = curve.Axis1;
+            Vector y = curve.Axis2;
+
+            Vector normal = new Vector { X = x.Y * y.Z - x.Z * y.Y, Y = x.Z * y.X - x.X * y.Z, Z = x.X * y.Y - x.Y * y.X };
+
+            return new Circle()
+            {
+                Centre = curve.Centre,
+                Radius = curve.Radius1,
+                Normal = normal,
+            };
+        }
+
+        /***************************************************/
+
+        public static explicit operator Circle(Arc curve)
+        {
+            if (System.Math.Abs((curve.EndAngle - curve.StartAngle) - System.Math.PI * 2) > Tolerance.Distance)
+                return null;
+
+            return new Circle()
+            {
+                Centre = curve.CoordinateSystem.Origin,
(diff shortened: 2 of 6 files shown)


Geometry_oM/Curve/Arc.cs
```diff
@@ -87,6 +88,44 @@ public class Arc : ICurve
...
 
         /***************************************************/
+
+        public static explicit operator Arc(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
```
pawelbaran, line 98: I would recommend changing to null checks on both `c` and output of `c.GetType()`.

By the way, why string matching, not type matching?

Geometry_oM/Curve/Arc.cs
```diff
@@ -87,6 +88,44 @@ public class Arc : ICurve
...
+            {
+                case "Arc":
+                    return c as Arc;
+                case "Circle":
+                    Circle circle = c as Circle;
```
pawelbaran, line 103: I think you can save some fraction of ms by `(Circle)c` instead of `c as Circle` - general rule for all casts you do here.

Geometry_oM/Curve/Circle.cs
```diff
@@ -42,6 +42,70 @@ public class Circle : ICurve
...
+        /**** Explicit Casting - Special Case           ****/
+        /***************************************************/
+
+        public static explicit operator Circle(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+            
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
```
pawelbaran, line 56: Same comments as for `Arc`

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -91,6 +91,42 @@ public class Ellipse : ICurve
...
 
         /***************************************************/
+
+        public static explicit operator Ellipse(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
```
pawelbaran, line 101: Same comments as for `Arc`

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -91,6 +91,42 @@ public class Ellipse : ICurve
...
+        }
+
+        /***************************************************/
+
+        public static explicit operator Ellipse(Arc curve)
+        {
+            Circle circle = (Circle)curve;
+            if (circle == null)
+                return null;
+            return (Ellipse)circle;
```
pawelbaran, line 126: Missing blank after `if` for readability

Geometry_oM/Curve/Line.cs
```diff
@@ -38,7 +38,56 @@ public class Line : ICurve
...
+        }
+
+        /***************************************************/
+
+        public static explicit operator Line(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
```
pawelbaran, line 72: Same comments as for `Arc`

Geometry_oM/Curve/NurbsCurve.cs
```diff
@@ -123,6 +123,38 @@ public class NurbsCurve : ICurve
 
         /***************************************************/
 
+        public static explicit operator NurbsCurve(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            switch (c.GetType().Name)
```
pawelbaran, line 132: Same comments as for `Arc`

Geometry_oM/Curve/Polyline.cs
```diff
@@ -46,6 +47,66 @@ public class Polyline : ICurve
...
+                }
+            }
+
+            Polyline result = new Polyline();
+
+            result.ControlPoints.Add(polyLines[0].ControlPoints[0]);
+            for (int i = 1; i < polyLines.Count; i++)
+            {
+                result.ControlPoints.AddRange(polyLines[i].ControlPoints.Skip(1));
+            }
```
pawelbaran, line 104: That is not safe - you need to make sure the `PolyCurve` is not disjoint (`Polyline` is continuous by definition)

Geometry_oM/Curve/Polyline.cs
```diff
@@ -46,6 +47,66 @@ public class Polyline : ICurve
...
+
+        public static explicit operator Polyline(PolyCurve polyCurve)
+        {
+            if (polyCurve.Curves.Count == 0)
+                return null;
+
+            List<Polyline> polyLines = new List<Polyline>();
+            foreach (ICurve c in polyCurve.Curves)
+            {
+                switch (c.GetType().Name)
```
pawelbaran, line 69: Again, you should `switch c.GetType() case typeof()` imo

Geometry_oM/Curve/Arc.cs
kThorsager: Unfortunally not as:

> Worth mentioning that the reason it got a bit messy in casting from PolyCurve to something is that C# does not support user defined casting from Interfaces. which means that these casts will not be that easily used in the code as one might hope.

`c` is a `ICurve` and the explicit cast can't be used unless it's from an object type.
I tried.

Would be happy if I was wrong tho

Geometry_oM/Curve/Polyline.cs
kThorsager: Thought `PolyCurve` was continuous. The more you know

Geometry_oM/Curve/Arc.cs
pawelbaran: Not sure if I get it - it compiles nicely after the change, does it go wrong runtime?

Geometry_oM/Curve/Polyline.cs
kThorsager: image
Not sure what you mean but does not seem to be working?

Geometry_oM/Curve/Arc.cs
kThorsager: oh yes, throws an error which stops the whole component

Geometry_oM/Curve/Polyline.cs
pawelbaran: Right, good point, this will not work. Could go for `if/else if (x is y)` as an alternative? Def faster (and more correct?), though not sure if there is no risk of this getting out of hand?

kThorsager: I'll give it a try, doesn't sound worse

Geometry_oM/Curve/Arc.cs
pawelbaran: That is unexpected, but makes sense after a second thought. Ignore this comment then

Geometry_oM/Curve/Polyline.cs
```diff
@@ -97,6 +98,11 @@ public class Polyline : ICurve
             result.ControlPoints.Add(polyLines[0].ControlPoints[0]);
             for (int i = 1; i < polyLines.Count; i++)
             {
+                // Ensure continious
+                Vector v = polyLines[i].ControlPoints[0] - polyLines[i - 1].ControlPoints.Last();
+                if (Math.Abs(v.X) > Tolerance.Distance || Math.Abs(v.Y) > Tolerance.Distance || Math.Abs(v.Z) > Tolerance.Distance)
```
kThorsager, line 103: Did the dirty of just checking if any of the orthogonal distances where grater that an tolerance instead of the distance. Think it should be fine.

pawelbaran: Personally I would go for a private square distance method - or would that not work?

kThorsager: Don't know, but think I'll just change to a in line solution.

kThorsager: Relizing now writing it that its actually shorter than what I wrote and the sqrt is not needed 🤦‍♂️

Geometry_oM/Curve/Arc.cs
```diff
@@ -87,6 +88,41 @@ public class Arc : ICurve
...
         /***************************************************/
+
+        public static explicit operator Arc(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            if (c == null)
+                return null;
```
pawelbaran, line 99: This is not needed I think

Geometry_oM/Curve/Circle.cs
```diff
@@ -42,6 +42,66 @@ public class Circle : ICurve
...
+        /***************************************************/
+
+        public static explicit operator Circle(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+            
+            ICurve c = curve.Curves[0];
+            if (c == null)
+                return null;
```
pawelbaran, line 57: This is not needed I think

Geometry_oM/Curve/Ellipse.cs
```diff
@@ -91,6 +91,41 @@ public class Ellipse : ICurve
...
         /***************************************************/
+
+        public static explicit operator Ellipse(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            if (c == null)
+                return null;
```
pawelbaran, line 102: This is not needed I think

Geometry_oM/Curve/Line.cs
```diff
@@ -38,7 +38,53 @@ public class Line : ICurve
...
+            if (c == null)
+                return null;
```
pawelbaran, line 74: This is not needed I think

Geometry_oM/Curve/NurbsCurve.cs
```diff
@@ -123,6 +123,33 @@ public class NurbsCurve : ICurve
...
         /***************************************************/
 
+        public static explicit operator NurbsCurve(PolyCurve curve)
+        {
+            if (curve.Curves.Count != 1)
+                return null;
+
+            ICurve c = curve.Curves[0];
+            if (c == null)
+                return null;
```
pawelbaran, line 133: This is not needed I think

Geometry_oM/Curve/Polyline.cs
```diff
@@ -46,6 +48,68 @@ public class Polyline : ICurve
...
+        public static explicit operator Polyline(PolyCurve polyCurve)
+        {
+            if (polyCurve.Curves.Count == 0)
+                return null;
+
+            List<Polyline> polyLines = new List<Polyline>();
+            foreach (ICurve c in polyCurve.Curves)
+            {
+                if (c == null)
+                    return null;
```
pawelbaran, line 71: This is not needed I think

Geometry_oM/Curve/Polyline.cs
```diff
@@ -46,6 +48,68 @@ public class Polyline : ICurve
...
+                    else
+                        polyLines.Add(pl);
+                }
+                else
+                    return null;
+            }
+
+            Polyline result = new Polyline();
+
+            result.ControlPoints.Add(polyLines[0].ControlPoints[0]);
```
pawelbaran, line 98: Shouldn't this line copy all control point of the 1st polyline?

kThorsager: Your're completely right

Geometry_oM/Curve/Arc.cs
kThorsager: Right it'll just go down to the last else if it's `null`, neat

kThorsager:
Have some thoughts in the wiki: https://github.com/BHoM/BHoM/wiki/Geometry_oM:-Explicit-casts

Figuring that the ins and outs of `PolySurface`s could be done without to much hassle, but am unused to working with `ISurface`s and wanted to keep a reasonable scope

pawelbaran:
Yeah would def not push it in this PR, was more curious. Seems to be worth a shot too.

al-fisher:
Once again - agree with @IsakNaslundBh 😄

PR #861, opened 2020-05-13
Geometry_oM: Make PlanarSurface IImmutable

### LINKED ISSUE #676: Geometry_oM: Make PlanarSurface immutable
#### Description:

Following the discussion in [#1429](https://github.com/BHoM/BHoM_Engine/pull/1429), `PlanarSurface` should be immutable and `Create` should check planarity of the input.

kThorsager:

Closes #676 

### Test files

Would recommend general project scripts as that is where I would expect things to go wrong, but that's also kind of expected.

Think the installer as a test should cover most issues

And don't really know what kind of test would be added otherwise.

### DIFF
--- a/Architecture_oM/Elements/Floor.cs
+++ b/Architecture_oM/Elements/Floor.cs
@@ -40,7 +40,7 @@ namespace BH.oM.Architecture.Elements
         /***************************************************/
 
         public virtual Object2DProperties Properties { get; set; } = new Object2DProperties();
-        public virtual ISurface Surface { get; set; } = new PlanarSurface();
+        public virtual ISurface Surface { get; set; } = null;
 
         /***************************************************/
     }
--- a/Architecture_oM/Elements/Roof.cs
+++ b/Architecture_oM/Elements/Roof.cs
@@ -37,7 +37,7 @@ namespace BH.oM.Architecture.Elements
         /***************************************************/
 
         public virtual Object2DProperties Properties { get; set; } = new Object2DProperties();
-        public virtual ISurface Surface { get; set; } = new PlanarSurface();
+        public virtual ISurface Surface { get; set; } = null;
 
         /***************************************************/
     }
--- a/Architecture_oM/Elements/Wall.cs
+++ b/Architecture_oM/Elements/Wall.cs
@@ -37,7 +37,7 @@ namespace BH.oM.Architecture.Elements
         /***************************************************/
 
         public virtual Object2DProperties Properties { get; set; } = new Object2DProperties();
-        public virtual ISurface Surface { get; set; } = new PlanarSurface();
+        public virtual ISurface Surface { get; set; } = null;
 
         /***************************************************/
     }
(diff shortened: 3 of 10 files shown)


kThorsager:
> As far as I understood, the whole concept of making `PlanarSurface` immutable was to centralize planarity and self-intersection checks inside the constructor in order to lift this burden from individual methods. At the moment immutability seems not to bring any added value, from what I understand?

- It keeps people from setting the values as they please whenever, now requires one to re-create the object.
- This constructor is not available from the UIs
- This follows the same pattern as say, `SectionProperty` which also can be set however one likes with the constructor from the code but gets protected in the UI and so forth.
https://github.com/BHoM/BHoM/blob/76f4350cfa569b2c0488029309d339248f701ff1/Structure_oM/SectionProperties/AluminiumSection.cs#L133-L181

pawelbaran:
Fair enough, but in this case we need the checks to be added in `Create` method.

kThorsager:
> Fair enough, but in this case we need the checks to be added in `Create` method.

Closed-ness and Planarity is checked, but we don't have any methods for checking self-intersections and such which works for Nurbs, which is why I did not add that

IsakNaslundBh:
Agree with @kThorsager s points regarding the checking.
And to reinforce the point, generally no checking should be done on the constructor level, but all checking done in the create methods.
The checking should be aimed at helping users, but from the side of the code we need to trust developers to make the choice if they want to use the checking the create methods provide, or prefer to "risk it" and call the constructor directly. Same principle as all Immutable objects in Structure_oM as Kalle pointed out.

pawelbaran:
Yeah, but then to make this change sensible, necessary checks need to be done in `Create` method. What we are still missing there is (at least):
- checks if all internal outlines lie inside the external outlines
- self-intersection checks

Without the above the change in oM is not meaningful as the user is still able to create invalid objects. Maybe we could add whatever needed with nurbs being skipped with a warning highlighting that fact?

IsakNaslundBh:
> Yeah, but then to make this change sensible, necessary checks need to be done in `Create` method. What we are still missing there is (at least):
> 
> * checks if all internal outlines lie inside the external outlines
> * self-intersection checks
> 
> Without the above the change in oM is not meaningful as the user is still able to create invalid objects. Maybe we could add whatever needed with nurbs being skipped with a warning highlighting that fact?

Ofc we need to ensure that the create methods are preforming all necessary checks, but that is a different issue from this. All this is meant to do is to stop it being possible to do something like this:

image

pawelbaran:
> Ofc we need to ensure that the create methods are preforming all necessary checks, but that is a different issue from this.

Yes and no - having accepted your explanation on why it is preferred not to put checks in the oM, I would still prefer to get the whole job done (oM + checks in Engine) in one go instead of only half of it (oM only).

Happy to move the discussion to BHoM_Engine, if you think this makes a difference.

kThorsager:
Added checks on the Engine branch https://github.com/BHoM/BHoM_Engine/pull/1761

IsakNaslundBh:
> Happy to move the discussion to BHoM_Engine, if you think this makes a difference.

Sorry for being obstinate, but I kind of think it does 😄 .

Again, never disagreed with any of your points whatsoever, just if they are an issue for the oM 😄

PR #865, opened 2020-05-13
Environment_oM: Add AnalysisGrid and Mesh Results

### LINKED ISSUE #864: Environment_oM: AnalysisGrid object
Following discussion with @tg359 we want to have an AnalysisGrid object which can be handled by multiple toolkits.

The AnalysisGrid will be made up of:

 - List<IAnalyticalPoint> Points
 - External boundary curve
 - Internal opening/boundary curves
 - Hour Index (representing 0-8760 hours of the analytical year)

IAnalyticalPoint would be an interface which provides a contract for:
 - Geometry Point
 - double Value

Use of IAnalyticalPoint interface allows each toolkit to define their own AnalyticalPoint with additional properties as required.

@IsakNaslundBh would something like this be at all worthwhile to you in structures (therefore arguing for moving this to Analytical_oM instead) or not? If not then I'll keep it in Environment_oM.

FraserGreenroyd:
Fixes #864 

### Additional comments
Also fixed compliance of CSProj in this PR.

### DIFF
--- a/Environment_oM/Analysis/AnalysisGrid.cs
+++ b/Environment_oM/Analysis/AnalysisGrid.cs
@@ -27,10 +27,37 @@ using System.Text;
 using System.Threading.Tasks;
 
 using BH.oM.Base;
+using BH.oM.Geometry;
+
+using System.ComponentModel;
 
 namespace BH.oM.Environment.Analysis
 {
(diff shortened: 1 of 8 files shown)


FraserGreenroyd:
@adecler @al-fisher this is ready for merge if CheckInstaller comes back clean :smile:

PR #879, opened 2020-05-20
Structure_oM: Override name property for structural properties

### LINKED ISSUE #842: Structure_oM: Override Name and add description for all StructuralProperties
All structural properties should have the Name Property overridden, as it is a core property for those classes.

IsakNaslundBh:

Closes #842 

Overriding name for objects where it should be seen as a primary property, and adding description.


rwemay:
> Looking good. And happy to be merged - but perhaps a quick comment/question first in case further clarity could be helpful for user
> 
> i.e. is uniqueness required for this name for pushing? Naturally in the BHoM, we do not impose this. Is simply a human readable string. But certain structural analysis packages are relying on this, right?
> 
> perhaps :
> 
> `"Name of the Constraint. Assigning unique Names is required for various structural packages to be able to create the object."`
> 
> @IsakNaslundBh what do you think?

Agree - will be really helpful, I might have said 'Note that assigning unique Names of properties is required for some structural packages.'

JosefTaylor:
to further bike-shed, since it's only required for some software, I'd lead with that:
"Note that for some structural packages, these Names must be unique"

IsakNaslundBh:
Agree, good comment! Now just need to pick my favourite of the three wording suggestions 😃

IsakNaslundBh:
Seem to be some problems with deserialisation after overriding the name, for section in particular. Could come from immutable obejcts. Putting do-not-merge tag until resolved.

Issue raised here:
https://github.com/BHoM/BHoM_Engine/issues/1796

IsakNaslundBh:
After chat with @al-fisher , updated the description on all of the properties for all to the same, including the "Unique" keyword.

IsakNaslundBh:
Serialiser_Engine is now merged, so this should be good to go now. Needs an approving review @al-fisher @peterjamesnugent @JosefTaylor

Structure_oM/Constraints/BarRelease.cs
```diff
@@ -33,6 +33,9 @@ public class BarRelease : BHoMObject, IProperty
         /**** Properties                                ****/
         /***************************************************/
 
+        [Description("Unique Names are required for some structural packages to create and identify the object.")]
```
peterjamesnugent, line 36: Should it not be:
A unique name is required for some structural packages to create and identify the object.

As the description is applying to a single object and only one name is required.

IsakNaslundBh: Could get behind that, yes

peterjamesnugent: Also, I wonder if name should be capitalised, given that it's referring to a property in an object it should be right? That was the convention we followed.

IsakNaslundBh:
@al-fisher @rwemay this should be good for merging

PR #890, opened 2020-05-25
Data_oM: LogicalNotRequest implemented

### LINKED ISSUE #889: Data_oM: Implement LogicalNotRequest
#### Description:

Following @alelom's request, it would be nice to have an IRequest that would invert the original query, e.g. to pull everything that is not a beam instead of all beams.

pawelbaran:

Closes #889

### Test files

See https://github.com/BHoM/Revit_Toolkit/pull/731

### Changelog

- `LogicalNotRequest` has been implemented
- `Requests` property has been removed from `ILogicalRequest`

### Additional comments

- I am not entirely happy about the name of the `Request` property (@IsakNaslundBh neither) - @al-fisher @rwemay do you have any native-speaker suggestions on that one?
- removal of property from an interface should not introduce any versioning issues, as long as the code complies everywhere, right @adecler?

### DIFF
--- a/Data_oM/Data_oM.csproj
+++ b/Data_oM/Data_oM.csproj
@@ -64,6 +64,7 @@
     <Compile Include="Collections\VennDiagram.cs" />
     <Compile Include="Requests\IResultRequest.cs" />
     <Compile Include="Requests\LogicalAndRequest.cs" />
+    <Compile Include="Requests\LogicalNotRequest.cs" />
     <Compile Include="Requests\LogicalOrRequest.cs" />
     <Compile Include="Requests\SelectionRequest.cs" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


alelom:
@pawelbaran regarding the naming, `LogicalNotRequest` makes total sense to me because it's consistent with the `LogicalAndRequest` etc.

alelom:
> @pawelbaran regarding the naming, `LogicalNotRequest` makes total sense to me because it's consistent with the `LogicalAndRequest` etc.

That said, this is one of those cases where I feel that capitalisation could help, but that should be done for all requests of this kind:
`LogicalNOTrequest` `LogicalANDrequest` etc.

pawelbaran:
Thanks @alelom. I was more concerned about the name of the property, being a generic `Request` atm.

al-fisher:
> That said, this is one of those cases where I feel that capitalisation could help, but that should be done for all requests of this kind:
LogicalNOTrequest LogicalANDrequest etc.

I actually think this might be good idea @alelom @pawelbaran - but would need consideration across all logical requests as you mention - so happy to merge this and an issue raised for discussion

PR #892, opened 2020-05-26
Reflection_oM: Update output to allow larger number of outputs

### LINKED ISSUE #891: Reflection_oM: Add Output<...> supporting larger number of outputs
#### Description:

Current limit is 5. A few methods will need larger number of outputs than the current limit. Should add up to 10.

rwemay:

Closes #891

### DIFF
--- a/Reflection_oM/Output.cs
+++ b/Reflection_oM/Output.cs
@@ -86,5 +86,110 @@ namespace BH.oM.Reflection
     }
 
     /***************************************************/
+
+    public class Output<T1, T2, T3, T4, T5, T6> : IObject, IOutput
+    {
+        public virtual T1 Item1 { get; set; }
+
+        public virtual T2 Item2 { get; set; }
+
+        public virtual T3 Item3 { get; set; }
+
+        public virtual T4 Item4 { get; set; }
+
+        public virtual T5 Item5 { get; set; }
+
+        public virtual T6 Item6 { get; set; }
+    }
+
+    /***************************************************/
+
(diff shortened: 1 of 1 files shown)


rwemay:
This is to facilitate a method that's returning 7 outputs but has scope to increase. @IsakNaslundBh  and I thought 10 as a practical limit.

rwemay:
@IsakNaslundBh , I think may be better to go with my initial gut of a dictionary, although not just for 'other' but all objects output. What do you think?

FraserGreenroyd:
> This is to facilitate a method that's returning 7 outputs but has scope to increase. @IsakNaslundBh and I thought 10 as a practical limit.

Ok that seems reasonable to me :smile:

al-fisher:
> @IsakNaslundBh , I think may be better to go with my initial gut of a dictionary, although not just for 'other' but all objects output. What do you think?

Yeah - not ideal to have to make the choice of when to output dictionary or when to manage with multi-output. However 10 will also not be enough! We will always want more 😄 
So inclined to agree with a dictionary approach

rwemay:
@FraserGreenroyd , I also thought about breaking into smaller parts, but for this particular method it does need a significant number of outputs as we're deriving things from a significant number of objects/relationships.

adecler:
> However 10 will also not be enough! We will always want more

I am not sure I agree. There is a point where the outputs make more sense as a single bundle object (e.g. `IResult`). 

Also worth mentioning that methods requiring a larger number of outputs will be exponentially rarer. And we have only needed 5 outputs so far despite having thousands of methods. So not convinced we will ever reach more than 10 that cannot be logically bundled.

PR #894, opened 2020-05-26
Reflection_oM: Adding the IOutputObejct interface

### LINKED ISSUE #893: Reflection_oM: Add new interface for defining objects to be automatically exploded and never show up in the UI
#### Description:

To more easily handle methods returning multiple outputs, in a more manageable format for the toolkit, suggesting adding a new interface that can be attached to an object as `IOutputObject`. 

Any method returning this object should make it automatically explode in the UI, and display all of its properties, including their descriptions.

Any object implementing this interface should be filtered out from autocontructor and type methods in the UI.

IsakNaslundBh:

Closes #893

- Adding interface to be applied to objects that when returned from a method should be automatically exploded.

### DIFF
--- a/Reflection_oM/Interface/IOutputObject.cs
+++ b/Reflection_oM/Interface/IOutputObject.cs
@@ -24,7 +24,7 @@ using System.ComponentModel;
 
 namespace BH.oM.Reflection.Interface
 {
-    [Description("Interface to be attached to obejcts that are to be automatically exploded when returned in any UI. Objects implementing this interface will not be constructable from the UI.")]
+    [Description("Interface to be attached to objects that are to be automatically exploded when returned in any UI. Objects implementing this interface will not be constructable from the UI.")]
     public interface IOutputObject
     {
     }
--- a/Reflection_oM/Interface/IOutputObject.cs
+++ b/Reflection_oM/Interface/IOutputObject.cs
@@ -1,0 +1,31 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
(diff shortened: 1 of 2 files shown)


Reflection_oM/Interface/IOutputObject.cs
```diff
@@ -0,0 +1,31 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System.ComponentModel;
+
+namespace BH.oM.Reflection.Interface
+{
+    [Description("Interface to be attached to obejcts that are to be automatically exploded when returned in any UI. Objects implementing this interface will not be constructable from the UI.")]
```
FraserGreenroyd, line 27: Sp. `obejcts` should be `objects`.

IsakNaslundBh:
> Is this instead of #892 ? Or as well as?

Could be both, but for the use case needed, would be instead. This is mostly a prototype to test things out. Will add appropriate tags.

FraserGreenroyd:
> > Is this instead of #892 ? Or as well as?
> 
> Could be both, but for the use case needed, would be instead. This is mostly a prototype to test things out. Will add appropriate tags.

Ok makes sense. Yeah probably good to be WIP cause I can see a few framework related questions arising out of this :smile:

IsakNaslundBh:
> > > Is this instead of #892 ? Or as well as?
> > 
> > 
> > Could be both, but for the use case needed, would be instead. This is mostly a prototype to test things out. Will add appropriate tags.
> 
> Ok makes sense. Yeah probably good to be WIP cause I can see a few framework related questions arising out of this 😄

yep, agreed. Have tried tackle them in the adjoining PRs in the engine and UI, but not sure it is the most elegant/best way to handle it.

adecler:
I have to admit I am a bit unsure about this one. I understand the idea behind it but I wonder if it isn't still better to output a single object that can then be exploded. We can for example make the analogy to the `IResult` objects. Those could easily be considered as `IOutputObject` and I would much prefer to keep them as they are.

IsakNaslundBh:
Closing this PR as development and discussions have moved on. Leaving issue open, with a parked label on it for further reference.

adecler:
@IsakNaslundBh , was that intentional that you didn't delete the branch ?

IsakNaslundBh:
Was not, thanks!

PR #916, opened 2020-06-10
LifeCycleAssessment_oM: Add LifeCycleAssessment results objects for element and project results

### LINKED ISSUE #915: LifeCycleAssessment_oM: Add LCA Results objects
#### Description:
Add objects for LCA results. This will require a per-object result object containing result value, result field type (GWP, acidification, etc), and additional scope and assumption (ie epd used and LCA stages) metadata. It will also require a per project result containing a collection of per object results with project metadata.

enarhi:

Closes #915 

Created objects for LCA results that standardize the recording and collection of LCA results on a per element and per project basis.


LifeCycleAssessment_oM/Enums/ObjectCategory.cs
```diff
@@ -0,0 +1,53 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.LifeCycleAssessment
+{
+    public enum ObjectCategory
+    {
+        Slab,
```
FraserGreenroyd, line 27: Enum should be alphabetised, but with `Undefined` as the first option - this is the order it renders in the UI and is easier for maintenance :smile:

LifeCycleAssessment_oM/Enums/ObjectScope.cs
```diff
@@ -0,0 +1,34 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+namespace BH.oM.LifeCycleAssessment
+{
+    public enum ObjectScope
+    {
+        Enclosure,
```
FraserGreenroyd, line 27: As above, alphabetised and with `Undefined` as first option

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+using BH.oM.Analytical.Results;
+using BH.oM.LifeCycleAssessment.MaterialFragments;
+using BH.oM.Geometry;
+using System.ComponentModel;
+using BH.oM.Base;
+using System;
+
+namespace BH.oM.LifeCycleAssessment.Results
+{
+    [Description("Result class for a LifeCycleAssessment of a single object")]
```
FraserGreenroyd, line 32: Might be worth a longer description to describe how this can be used/might store - but we can handle that after this PR if necessary.

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+    public class LifeCycleAssessmentElementResult : IResult, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Id of the BHoMObject that this result belongs to.")]
+        public virtual IComparable ObjectId { get; } = "";
+
+        [Description("Identifier for the case evaluated, ie GlobalWarmingPotential or Acidification")]
```
FraserGreenroyd, line 42: Should be `e.g.` unless those are the only two options that it can be

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+        public virtual IComparable ResultCase { get; } = "";
+
+        [Description("Time step for time history results (This is unlikely for LCA).")]
+        public virtual double TimeStep { get; } = 0.0;
+
+        [Description("Scope the object this result was generated from belongs to, ie Foundation or Facade")]
```
FraserGreenroyd, line 48: See above RE: `e.g.`

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+        public virtual ObjectScope Scope { get; } = ObjectScope.Undefined;
+
+        [Description("Category of the object this result was generated from, ie Beam or Wall")]
```
FraserGreenroyd, line 51: See above RE: `e.g.`

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+        public virtual ObjectCategory Category { get; } = ObjectCategory.Undefined;
+
+        [Description("Quantity of metric evaluated per the ResultCase, ie total kgCO2e for GlobalWarmingPotential.")]
```
FraserGreenroyd, line 54: See above RE: `e.g.`

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+        public virtual double Quantity { get; } = 0.0;
+
+        [Description("The EnvironmentalProductDeclaration used to generate this result.")]
+        public virtual IEnvironmentalProductDeclarationData EnvironmentalProductDeclaration{ get; }
```
FraserGreenroyd, line 58: Given we have the ObjectID, how necessary is this property?

LifeCycleAssessment_oM/Results/LifeCycleAssessmentResult.cs
```diff
@@ -0,0 +1,100 @@
...
+    public class LifeCycleAssessmentResult : IResult, IResultCollection<LifeCycleAssessmentElementResult>, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("Id of the project that this result is for.")]
+        public virtual IComparable ObjectId { get; } = "";
+
+        [Description("Identifier for the case evaluated, ie GlobalWarmingPotential or Acidification")]
```
FraserGreenroyd, line 43: See above RE: `e.g.`

LifeCycleAssessment_oM/Results/LifeCycleAssessmentResult.cs
```diff
@@ -0,0 +1,100 @@
...
+using BH.oM.LifeCycleAssessment.MaterialFragments;
+using BH.oM.Geometry;
+using System.ComponentModel;
+using BH.oM.Base;
+using System;
+using System.Collections.ObjectModel;
+
+namespace BH.oM.LifeCycleAssessment.Results
+{
+    [Description("Result class for a LifeCycleAssessment of a whole project")]
```
FraserGreenroyd, line 33: See above RE: description

LifeCycleAssessment_oM/Results/LifeCycleAssessmentResult.cs
```diff
@@ -0,0 +1,100 @@
...
+        public virtual IComparable ObjectId { get; } = "";
+
+        [Description("Identifier for the case evaluated, ie GlobalWarmingPotential or Acidification")]
+        public virtual IComparable ResultCase { get; } = "";
+
+        [Description("Time step for time history results (This is unlikely for LCA).")]
+        public virtual double TimeStep { get; } = 0.0;
+
+        [Description("A LifeCycleAssessmentScope object containing project information for the LifeCycleAssessment")]
+        public virtual LifeCycleAssessmentScope LCAScope { get; }
```
FraserGreenroyd, line 50: Given we have the ObjectID how necessary is this property?
If the property is necessary, then it shouldn't have an acronym in its property name.

LifeCycleAssessment_oM/Results/LifeCycleAssessmentResult.cs
```diff
@@ -0,0 +1,100 @@
...
+            LCAScope = lcaScope;
+            Results = results;
+            TotalGlobalWarmingPotential = globalWarmingPotential;
+        }
+
+        /***************************************************/
+        /**** IComparable Interface                     ****/
+        /***************************************************/
+
+        public int CompareTo(IResult other)
```
FraserGreenroyd, line 77: Why does this not get a description and the other one did? This one feels left out 😢

LifeCycleAssessment_oM/Results/LifeCycleAssessmentElementResult.cs
```diff
@@ -0,0 +1,116 @@
...
+        public virtual double TimeStep { get; } = 0.0;
+
+        [Description("Scope the object this result was generated from belongs to, ie Foundation or Facade")]
+        public virtual ObjectScope Scope { get; } = ObjectScope.Undefined;
+
+        [Description("Category of the object this result was generated from, ie Beam or Wall")]
+        public virtual ObjectCategory Category { get; } = ObjectCategory.Undefined;
+
+        [Description("Quantity of metric evaluated per the ResultCase, ie total kgCO2e for GlobalWarmingPotential.")]
+        public virtual double Quantity { get; } = 0.0;
```
IsakNaslundBh, line 55: My biggest comment would be regarding this property.

What we have done in structures at least is to create separate result classes for separate results, where it is then possible to hardcode the actual quantites, and give them a quntity attribute as well as a description.

This would mean for you case that you would have a subclass called for example `GlobalWarmingPotentialResult : LifeCycleAssessmentElementResult` that had only this property assigned to it, and then expanding to how many different result classes you would need for different results.

See for example `NodeReaction` vs `NodeDisplacement`.

As an alternative to having multiple classes, would be to add in all the various properties that you might expect to get back, and store them all under one result. This would also be what we have in structure where a `NodeReaction` has 3 force values as well as 3 moment values.

Basically, adding things like `GlobalWarmingPotential`, `OzoneDepletionPotential` etc as separate properties all in one class.

But think it is critical that a property can only be of one particular quantity/type for it to be crystal clear what it actually is, to a user being calculating the results, or pulling from a DB and comparing with something else etc.

LifeCycleAssessment_oM/Results/GlobalWarmingPotentialResult.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    [Description("Gives the total GlobalWarmingPotential of an object based on its EnvironmentalProductDeclaration.")]
+    public class GlobalWarmingPotentialResult : LifeCycleAssessmentElementResult, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("The total embodied carbon of the object in kgCO2e.")]
+        public virtual double GlobalWarmingPotential{ get; }
```
IsakNaslundBh, line 40: Could add a `[Mass]` quantity attribute to this one.

FraserGreenroyd: I second this

enarhi:
@IsakNaslundBh @FraserGreenroyd Mass attribute added.

FraserGreenroyd:
@al-fisher @adecler @rwemay this is ready for merge please ideally today as it's a feature! :smile:

PR #919, opened 2020-06-12
LifeCycleAssessment_oM: Fixed Scope Objects 

### LINKED ISSUE #915: LifeCycleAssessment_oM: Add LCA Results objects

enarhi:

Closes #915 (Already closed, follow up fix)

LCA Scopeobjects take a list of IBHoMObjects instead of IElementM's in order to allow for EPD calculation on non-IElementM objects.


FraserGreenroyd:
@al-fisher @adecler this is ready for merge if checks are happy please :smile:

PR #930, opened 2020-06-22
LifeCycleAssessment_oM: Update ProjectArea Enum

### LINKED ISSUE #929: Mismatch in LifeCycleAssessmentScope Language
#### Description:
Not exactly a bug but just a mismatch in this component between the "ProjectArea" which is accepted a category derived from the CLF benchmarking categories and is technically based on ft2, and the "ProjectActualArea" which is seeking a m2. 

I think these fields should probably agree with each other so perhaps we just revise the accepted categories to be in m2 (per the alternative snip), the native CLF dataset that would have the full list of alternative taxonomy for the m2 categories : https://burohappold.sharepoint.com/:f:/r/sites/LifeCycleAssessment/Best%20Practices%20Library/Resources%20and%20Research/Tools%20and%20Datasets/CLF-Embodied-Carbon-Benchmark-Report-and-Data?csf=1&web=1&e=WUHj4z

See image below documenting this: 
image

image

michaelhoehn:

Closes #929 

### Additional comments

Enum options reflect Carbon Leadership Forum baseline categories for m2. Previous versions were in ft and could cause confusion if utilising the benchmark functionality within the toolkit. 

For more details, please visit the [CLF Data Visualisation ](http://carbonleadershipforum.org/projects/embodied-carbon-benchmark-study-data-visualization/)

### DIFF
--- a/LifeCycleAssessment_oM/Enums/ProjectArea.cs
+++ b/LifeCycleAssessment_oM/Enums/ProjectArea.cs
@@ -20,18 +20,23 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using System.ComponentModel;
+
 namespace BH.oM.LifeCycleAssessment
 {
+    [Description("ProjectArea in m2 for Life Cycle Assessment comparison.")]
     public enum ProjectArea
     {
         Undefined,
-        LessThan1000,
-        Between1000and2500,
-        Between2500and5000,
-        Between5000and10000,
-        Between10000and20000,
-        Between20000and50000,
-        Between50000and100000,
-        GreaterThan100000,
+        Under93,
+        Between94and465,
+        Between466and929,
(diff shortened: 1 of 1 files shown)


LifeCycleAssessment_oM/Enums/ProjectArea.cs
```diff
@@ -20,18 +20,23 @@
...
+        Under93,
+        Between94and465,
+        Between466and929,
+        Between930and2323,
+        Between2324and4645,
+        Between4646and9290,
+        Between9291and18580,
+        Between18581and46451,
+        Between46452and92903,
+        Over92903,
```
FraserGreenroyd, line 40: What happens if a value is exactly 93? It is not under 93 and it is not between 94 and 465.

michaelhoehn: 42

PR #958, opened 2020-08-03
Inspection_oM: Add initial Inspection objects for Audits and Issues

### LINKED ISSUE #956: Create Inspection_oM
#### Description:

@enarhi 

In order to generalise the concept of Audit/Issues for IAuditor and 3DRepo.

### LINKED ISSUE #957: Inspection_oM: Create an Audit object

alelom:

Closes #957 
Closes #956


enarhi (dismissed):
LGTM. Alessio, please confirm this good to go on your end. Suggest we open separate issues as required for further object development to address specific 3D Repo workflow bugs.

FraserGreenroyd (changes requested):
I'm blocking because there is an issue with CheckInstaller which needs discussion before this can be merged, though I fear iAuditor_Toolkit might already be broken looking at the logs.

Also, a lot of the descriptions need improving, they're not really suitable for oM object properties, and a lot of the objects need defaults assigning to them.

The PR title is also incorrect, it should be `Inspection_oM: Add objects` and specifically maybe which objects to the change log or the title.

The branch name also contains a special character which the CI generally doesn't like and shouldn't be followed in the future please :smile:

enarhi:
CheckInstaller issues have been resolved. After discussing with @FraserGreenroyd I will take a pass at updating / improving descriptions and defaults, and we will raise follow-up issues as needed for further improvements to the descriptions after this pr is merged.

Inspection_oM/Audit.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        [Description("Job leader of the Project the Audit is for")]
+        public virtual string JobLeader { get; set; } = "";
+
+        [Description("List of people the Audit is distributed to")]
+        public virtual List<string> Distribution { get; set; } = new List<string>();
+
+        [Description("List of people in attendance during Audit")]
+        public virtual List<string> Attendance { get; set; } = new List<string>();
+
```
FraserGreenroyd, line 84: Thinking long term, it would be good to make use of the Humans_oM object for Humans as a nice way to utilise more aspects of the BHoM. Happy with strings for now, but just something to consider? :smile:

Inspection_oM/Audit.cs
```diff
@@ -0,0 +1,102 @@
...
+        [Description("Reason(s) for visit and Audit")]
```
FraserGreenroyd, line 85: So, in this description, `audit` refers to the event, rather than the object, correct? In which case it should be `audit` rather than `audit`. This is different when referring to the object itself (e.g. in the `Title of the Audit` which is referring to the definite article which is this object).

It's a minor issue, can we maybe do an issue to resolve in the future to avoid blocking this PR?

Inspection_oM/Comment.cs
```diff
@@ -0,0 +1,51 @@
...
+        /***************************************************/
+
+        [Description("Comment message")]
+        public virtual string Message { get; set; } = "";
+
+        [Description("Name of the individual making the comment")]
+        public virtual string Owner { get; set; } = "";
+
+        [Description("Date the comment was made")]
+        public virtual string CommentDate { get; set; } = "";
```
FraserGreenroyd, line 47: Why not a date time? This oM strays dangerously into being stringly typed!

Inspection_oM/Issue.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Inspection
+{
+    [Description("An issue belonging to an audit")]
+    public class Issue : BHoMObject
```
FraserGreenroyd, line 35: Is `Issue` the correct name of the object? I fear perhaps conflict with Programming Issues, or other types of Issues. Obviously context is important, and if the formal name during an audit is 'issue' then I have no qualms, so just checking if this is the formal name, or a colloquial name we're using?

Inspection_oM/Issue.cs
```diff
@@ -0,0 +1,74 @@
...
+        public virtual DateTime DateCreated { get; set; } = new DateTime();
+
+        [Description("Priority tag to better categorize your issue  identifying its urgency")]
+        public virtual string Priority { get; set; } = "";
+
+        [Description("Status description of the issue (eg open, in progress, closed)")]
+        public virtual string Status { get; set; } = "";
+
+        [Description("A list of assignees for the issue. These are the individuals / companies responsible for resolving the issue.")]
+        public virtual string Assign { get; set; } = "";
```
FraserGreenroyd, line 54: Description says list but property is just a string?

Inspection_oM/Issue.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [Description("Issue type (eg For Information)")]
+        public virtual string Type { get; set; } = "unassigned";
```
FraserGreenroyd, line 57: This should probably be an enum? Maybe an issue for later?

Inspection_oM/Issue.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [Description("The scope and nature of the issue")]
+        public virtual string Description { get; set; } = "";
+
+        [Description("Filenames of the images associated with the issue." +
+            "These must include the file extension.")]
+        public virtual List<string> Media { get; set; } = new List<string>();
+
+        [Description("Location of the issue in 3D space in terms of the project's coordinates")]
+        public virtual Point Position { get; set; } = new Point();
```
FraserGreenroyd, line 67: Coordinates in what coordinate system?

FraserGreenroyd (changes requested):
Some comments that don't need addressing, one that definitely does. Happy for discussion on the others or issues being raised to solve later. Otherwise is looking good @enarhi @alelom :smile:

Inspection_oM/Audit.cs
enarhi: Agreed. Strings for now but can move towards Humans in the future.

Inspection_oM/Issue.cs
enarhi: Will be changed to an enum in the future, we have not nailed down what that enum is limited to yet though so leaving as string for now.

Inspection_oM/Issue.cs
enarhi: Updated to specify project coordinate system, does that suffice?

Inspection_oM/Issue.cs
enarhi: Fixed.

Inspection_oM/Issue.cs
enarhi: Issue is the correct name based on discussion with those doing the inspections. I think it's appropriate left as is.

Inspection_oM/Comment.cs
enarhi: Fixed.

FraserGreenroyd (approved):
Happy with the responses and latest changes, thanks @enarhi :smile:

PR #984, opened 2020-09-04
Add Analytical Graph. Deprecate Graph, GraphNode and GraphLink

### LINKED ISSUE #983: Migration of Graph from Data_oM to Analytical_oM
Following on from discussion with @adecler and @al-fisher. Capturing first step in migrating prototype graph operations in `GraphFlow_Toolkit`. 

Migrate `Graph<T>` from Data_oM to Analytical_oM
Deprecate `Graph<T>`
Redefine as `Graph<TNode>` (or `Network` thoughts?) implement `where` clause to enforce use of `INode`. With properties `List<INode>` and `List<ILink>`.

Proposing that the Value property of the old `GraphNode<T>`  is redundant maybe even confusing. As following the `GraphNode<T>` implementation we can end up with something like:

```C#
//flow node implements INode
FlowNode flownode = new FlowNode();
GraphNode<FlowNode> graphnode = Create.GraphNode(flownode);
//access flownode properties on the graphnode properties is via Value:
double flowrate = graphnode.Value.FlowRate;
```
The attribute dictionary property of `GraphNode<T>` could be added as required on classes implementing `BH.oM.Analytical.Elements.INode`.

Similarly classes implementing `BH.oM.Analytical.Elements.ILink` would if needed include the weight and attribute dictionary properties on the `GraphLink<T>`.

This is the first step in enabling longer term possibilities including:
- defining and controlling watertight graph / network models within the BHoM
- allowing workflows spanning a continuum of relationships in networks
- working with possible decentralised `Relationship` fragments between any and all objects
- centralising graph operations and algorithms 
- more...

Short term this helps generalise development and integration GraphFlow_Toolkit into the BHoM 😸

rolyhudson:
Closes #983 
~~Labelled 'parked' as further discussion may be needed. Would like to get general consensus on change to structure of `Graph<T>`, `GraphNode<T>` and `GraphLink<T.>` that this is intended to replace.~~

### DIFF
--- a/Analytical_oM/Analytical_oM.csproj
+++ b/Analytical_oM/Analytical_oM.csproj
@@ -43,6 +43,7 @@
   <ItemGroup>
     <Compile Include="Elements\IEdge.cs" />
     <Compile Include="Elements\IFace.cs" />
+    <Compile Include="Elements\Graph.cs" />
     <Compile Include="Elements\ILink.cs" />
     <Compile Include="Elements\IMesh.cs" />
     <Compile Include="Elements\INode.cs" />
--- a/Analytical_oM/Elements/Graph.cs
+++ b/Analytical_oM/Elements/Graph.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * This file is part of the Buildings and Habitats object Model (BHoM)
  * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
  *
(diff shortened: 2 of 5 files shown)


al-fisher (commented):
One minor comment I have is around the namespace for the new `Graph` class.
This probably will want to sit somewhere other than `Elements` ultimately. As owns Elements rather than is one. I think.
Following recent discussions, best guess would be `Models` - with potentially new interfaces and similar topological entities in this namespace. Not sure though - and this naming could change. 
Can always tweak later though - so would not want to slow merging this either way.

al-fisher:
Or simply `Collections` similar to the previous Data_oM where we are migrating these concepts from. Either way - we (I) are likely to change minds so not a biggie

adecler:
@al-fisher , I am not sure about this one:
- An `IPanel` also contains others elements (`List<IEdge>` and `List<IOpening>`). So it might be a bit difficult to draw a clear line between what is a collection and what is an element simply based on the property types of a class. This is where engineering expertise comes into play I guess :wink:
- There is a possibly niche but interesting effect with the Graph being an IElement and therefore allowed to be contained in something bigger. There is also a recursive aspect to it that might be interesting. I have no idea if this is a pro or a cons though.

al-fisher:
Yes - fair point - this concept is not going to be so clear cut.
As still in development - very happy to leave namespace as is. Your point makes sense @adecler 
Am continuing sketches of the iModel workflow - can take as wholly separate exercise

al-fisher (approved):
As discussed - this looks good to me!

PR #1020, opened 2020-10-13
Reflection_oM: Adding attributes to better control auto-create components

### LINKED ISSUE #731: Control of oM properties as "Required" not possible through object initialiser auto-create
A disadvantage of the auto-generated object initializer create methods is the loss of control of whether a Property is _required_ or not when in the UI.

Would be good to control this from the oM itself as I think is a useful concept. And then reflect though.

#### _One idea would be to flag as a _required_ property by _NOT_ providing a default value?_

Obviously not too dissimilar to the way any C# method (and any of our Create methods using derived properties) is working.
Is there a disadvantage to that?   

As discussed with @IsakNaslundBh could potentially be quite neat and clear.

Obviously if using objects in the code you would need to initialise object properties before using (especially in the case of lists - up until now have in all cases been providing a default empty `new List<>` )
I think this would be a good sacrifice in the name of clarity though. 

Example for `Structure.Panel` 

image

Another option could be to have a `RequiredAttribute` - although we have quite a few attributes already.

Thoughts? @adecler @IsakNaslundBh @FraserGreenroyd all

### LINKED ISSUE #883: Reflection_oM: Add attribute to prevent auto-constructor method in the UI
#### Description:

Add attribute to supress the generation of the auto-created property assignemnt methods in the UI. Usefull when you have added an explicit create method that takes the core properties of the objects, but does additional checks, such as assigning default values and/or preforming additional checks on the incoming arguments.

adecler:

Closes #731
Closes #883 

Also helps with https://github.com/BHoM/BHoM_UI/issues/205

This provides the 3 attributes we agreed on when talking about improving the control of the auto-constructor components. 
- Required properties
- Classes that should not have an auto-constructor (because they have a Create method in the engine and the two cannot coexist).
- Warning on the component when the default value of a property is used (spoiler alter, I actually intend to make it a not and not a warning to help OCD people like me that cannot stand for a component to stay orange).

### Test files
Not much to test for now. It is more a matter of agreeing with the definition I created here.

### Additional comments
I am aware that there is plan to migrate the attributes from the reflection engine to the base engine (see https://github.com/BHoM/BHoM_Engine/issues/1645). So I could have create those new attributes in the base already but decided not to in order to keep things tidy.

### DIFF
--- a/Reflection_oM/Attributes/DefaultValueWarningAttribute.cs
+++ b/Reflection_oM/Attributes/DefaultValueWarningAttribute.cs
@@ -1,0 +1,57 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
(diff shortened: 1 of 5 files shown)


adecler:
> Think some description on the these attributes would be good though to help devs understand what they do and how they should be used.

Done

IsakNaslundBh (approved):
LGTM now!

PR #1032, opened 2020-10-16
Architecture_oM: Add ceiling tile object

### LINKED ISSUE #1029: Architecture_oM: Add ceiling tile object
To facilitate with https://github.com/BHoM/BHoM_Engine/issues/2049 we want a `CeilingTile` object. This will also help with the removal/renaming of CustomData as Revit_TK stores the ceiling pattern as a custom data parameter.

It would be good if the Ceiling Tile object can be made up of Edges (implementing IEdge from Analytical oM) so as to match other types of handling this. Maybe the Ceiling Tile can then be an IPanel? @IsakNaslundBh what do you think on that?

Alternatively, just a simple polyline defining its geometry is enough for now.

Then the Ceiling object can have a property that includes a collection of the `CeilingTile` objects which the Revit TK can then be updated to use instead of CustomData.

JohannaOlin1:
Closes #1029 

Facilitating BHoM/BHoM_Engine#2049 by creating a ceiling tile object.

### DIFF
--- a/Architecture_oM/Architecture_oM.csproj
+++ b/Architecture_oM/Architecture_oM.csproj
@@ -42,6 +42,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Elements\Ceiling.cs" />
+    <Compile Include="Elements\CeilingTile.cs" />
     <Compile Include="Elements\Grid.cs" />
     <Compile Include="Elements\Occupancy.cs" />
     <Compile Include="Elements\Room.cs" />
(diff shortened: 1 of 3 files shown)


FraserGreenroyd:
Have asked @IsakNaslundBh to check this doesn't conflict with him before merging happens :smile:

michaelhoehn (approved):
LGTM

adecler:
@FraserGreenroyd , was there any versioning provided separately to cover this change ?

FraserGreenroyd:
@adecler just to capture our chat from the call earlier for public record - no versioning was provided for this change as it was only a property addition :smile:

PR #1033, opened 2020-10-19
Introduce new AdapterIdFragment to replace CustomData usage in Adapter

### LINKED ISSUE #1012: Base_oM: create new IAdapterIdFragment interface
#### Description:

This is to store any adapter Id, so we can move them from being stored in the CustomData. 

See also https://github.com/BHoM/BHoM_Adapter/issues/256.

alelom:

Closes #1012


BHoM/AdapterIdFragment.cs
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+
+namespace BH.oM.Base
+{
+    public class AdapterIdFragment : IAdapterIdFragment
```
FraserGreenroyd, line 28: Can you explain this design choice? I'm not sure why we would need an object and an interface with almost the same name? I'm probably missing something so apologies for the potentially dumb question!

Also - can the PR title be updated to include which oM is modified.

alelom: Actually not needed - it was an initial test. Removed.

FraserGreenroyd (commented):
Apparently I have to post this comment to let my other comment be seen...

FraserGreenroyd:
And by the time I finish posting @alelom already changes is all - lovely jubbly 😄

BHoM/AdapterIdFragment.cs
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+
+namespace BH.oM.Base
+{
+    public class AdapterIdFragment : IFragment
```
al-fisher, line 28: Thanks @alelom  - great to get this in.

One comment - I might simply call this `AdapterID : IFragment` 
i.e. I think all objects implementing `IFragment` should not have to have the suffix `Fragment` in their name.

alelom: True, thanks. Also, I converted this to draft as after some chat with @IsakNaslundBh we went through another prototype.

al-fisher: Thanks @alelom 
OK - Perfect.

adecler:
1. Looking at the original commits. Find they make perfect sense (after the corrections from Al and Fraser were applied at least)
2. Notice another prototype incoming after conversation with Isak
3. Confused as to why we ended up with such a complex solution (other than the Isak effect of course)

😋 😉

IsakNaslundBh:
> Confused as to why we ended up with such a complex solution (other than the Isak effect of course)

Was an attempt to simplify Id retrieval further down the line. Ids that implement the generic interface will not need any additional methods etc in the toolkits. Alessio is still prototyping it out, if the generics end up looking as it is costing more in complexity than it gives I am fine with getting rid of it.

BHoM/Interface/IAdapterId.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace BH.oM.Base
+{
+    public interface IAdapterId : IFragment
+    {
+    }
+
+    public interface IAdapterId<T> : IAdapterId, IFragment
+    {
+        T Id { get; set; }
+    }
```
FraserGreenroyd, line 35: Sorry if this is in a discussion that I've missed somewhere, but why do we have both? It might be best if @IsakNaslundBh  reviews this PR cause he's probably been more involved in the discussion? Also, structures use the Adapter IDs more than Environments, so he's probably much better placed to review functionality? :smile:

IsakNaslundBh: Reason we tried with a generic and a non-generic is sort of the same reason we have `IElement1D` and `ILink`. The non-generic interface requires you to have some functionality. The generic gives you the base functionality for free. Basically, any class that implements IAdapterId will be required to have a method for getting and setting the Id. if you implement the generic interface, that method will be already solved for you in the adapter engine.

I know it is a bit of extra complication, but from what @alelom  has done so far, it seem to work quite well I think

alelom: > Sorry if this is in a discussion that I've missed somewhere, but why do we have both? It might be best if @IsakNaslundBh reviews this PR cause he's probably been more involved in the discussion? Also, structures use the Adapter IDs more than Environments, so he's probably much better placed to review functionality? 😄

First, @IsakNaslundBh is the very reason why we have both 😄 

I hate to defend generics – in fact my original proposal was relying on `System.Object` – however this does simplify things down the line in the adapters. 

In summary, the base interface gives you these advantages:
- As @IsakNaslundBh explained, the generic interface simplifies the code down the line. 
- the base interface also gives you flexibility down the line in case you don't plan on using the above 
- since we have the generic version, having a base interface in addition to the generic simplifies the code when Type checking (e.g. doing `.OfType<IAdapterId>()` would not be possible with the generic interface alone)

BHoM/Interface/IAdapterId.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        [Description("Identifier of the object in the external software.")]
+        object Id { get; set; }
+    }
+
+    public interface IPersistentAdapterId : IFragment
+    {
+        [Description("Globally unique and generated upon object creation in the external software; it never changes throughout the life of the object.")]
+        object PersistentId { get; set; }
+    }
```
IsakNaslundBh, line 39: Needs to be moved to a separate file

IsakNaslundBh (approved):
All looks good to me now!

IsakNaslundBh:
Updated this from master to be able to get relevant check installer checks as some breaking changes for LCA had been merged in fyi @alelom

PR #1074, opened 2020-11-10
Structure_oM: Add BarDifferentialTemperatureLoad and DifferentialTemperatureLoad

### LINKED ISSUE #945: Structure_oM: Make temperature loads able to handle non-uniform distribution
#### Description:

Current temperature loads only supports uniform temperature changes. Load need to updated to as a minimum support a linear temperature gradient throughout the elements, and look into possibility to support non-linear/free form distributions.

peterjamesnugent:

Closes #945

@IsakNaslundBh and @JosefTaylor some thoughts:
- I have opted to use a `Dictionary<double, double>` to represent the profiles, this is consistent with the codified approaches and if you want a nonlinear profile this can be achieved with small increments
- I have added a `Vector` to specify the load direction (I have caveated this in the description that most analytical software packages only accept local y and local z)
- I have removed the `LoadAxis` and `Projected` inputs from the `Create` method as they are not relevant, I was unable to remove from the object as they are required by `IElementLoad<T>`
- Perhaps we could add an `eNum` to limit the choice given that this is a load applied locally 
- This will require a PR on every structural toolkit but I think if we left it as `BarTemperatureLoad` and `AreaTemperatureLoad` it would cause confusion, this feels a lot clearer - happy to hear your thoughts
- Versioning appears to work for the types, namespaces and serialised objects which is good

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM/Structure_oM/Issue945-BarDifferentialLoad?csf=1&web=1&e=3BovsF

### Changelog

- `BarTemperatureLoad` renamed to `BarUniformTemperatureLoad` and `AreaTemperatureLoad` renamed to `AreaUniformTemperatureLoad`
- This is to reflect the addition of a `BarDifferentialTemperatureLoad` and `AreaDifferentialTemperatureLoad`
- The uniform temperature loads represent a constant global temperature applied to the entire element
- The differential temperature loads represent a difference in temperature between the faces of the profile
- The `BarDifferentialTemperatureLoad` can be applied in either the `local-x` or `local-y` 
- The `AreaDifferentialTemperatureLoad` is applied in the vertical direction only


IsakNaslundBh:
Think it generally looks good! Have a few comments, or discussion points to your points.

> * I have opted to use a `Dictionary<double, double>` to represent the profiles, this is consistent with the codified approaches and if you want a nonlinear profile this can be achieved with small increments

Ofc good that this is flexible and more future proof, but wonder if it is adding complication without gaining that much?

Have checked GSA, Robot and ETABS at least, and for those, GSA and Robot has support for a linear gradient. ETABS has no support for varying temperature loads at all.

Ofc, there are more packages than the ones stated above, so if we want to keep this for other reasons I think that is fine. Not against it in principle!

Can always have simpler create methods to guide users, like just providing the top and bottom tempratures as well as distances from the centre of the bar/panel.

> * I have added a `Vector` to specify the load direction (I have caveated this in the description that most analytical software packages only accept local y and local z)
> * Perhaps we could add an `eNum` to limit the choice given that this is a load applied locally

Think an enum to limit the choice to local y and z instead of a vector could work better.

> * I have removed the `LoadAxis` and `Projected` inputs from the `Create` method as they are not relevant, I was unable to remove from the object as they are required by `IElementLoad<T>`

yeah, might want to move those properties out from the base load, given there are a couple of loads they are not relevant for..

> * This will require a PR on every structural toolkit but I think if we left it as `BarTemperatureLoad` and `AreaTemperatureLoad` it would cause confusion, this feels a lot clearer - happy to hear your thoughts

Yeah, agree, even if it makes this more painful

> * Versioning appears to work for the types, namespaces and serialised objects which is good

👍

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check code compliance

IsakNaslundBh:
@BHoMBot check project compliance

IsakNaslundBh:
@BHoMBot check documentation compliance

IsakNaslundBh:
@BHoMBot check copyright compliance

IsakNaslundBh:
@peterjamesnugent I updated from master as the MEP changes completely destroyed the check installer

IsakNaslundBh:
@BHoMBot check project compliance

IsakNaslundBh:
@BHoMBot check code compliance

IsakNaslundBh:
@BHoMBot check documentation compliance

IsakNaslundBh:
@BHoMBot check copyright compliance

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check project compliance

IsakNaslundBh:
@BHoMBot check code compliance

IsakNaslundBh:
@BHoMBot check documentation compliance

IsakNaslundBh:
@BHoMBot check copyright compliance

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh (approved):
Happy with this! Approved!

PR #1075, opened 2020-11-11
Structure_oM: Updates to BarVaryingDistributedLoad

### LINKED ISSUE #507: Structure_oM: BarVaryingDistributedLoad review
The properties DistanceFromA and DistanceFromB needs a review.

Need to specify unit (paramter 0-1 or meters), currently it is assuming meters. This need to be put in writing. Also, review needed if having distance from start and end is the right way to do it or if both values should be in reference to the starting point.

IsakNaslundBh:
This needs to be coordinated with https://github.com/BHoM/BHoM/pull/1074 .

Raising this as draft and for discussion before all other changes in toolkits as well as versioning can be actioned!

Closes #507

### Changelog

- Rename the properties, using word Start and End instead of A and B
- Change to measure distance to both start and end from the start node instead of end being measured from the end node
- Adding property to control if the positions should be relative (0-1) or measured in meters.

### Additional comments

@peterjamesnugent @JosefTaylor @jdarcyBHE @rwemay @johannaisak glad for any initial comments on this!


jdarcyBHE:
Proposed changes make sense, possibly coming from a robot view but I think the default on the length type should be relative.

IsakNaslundBh:
Will test running bhombot check isntaller to check the impact on this organisation.

@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check installer

Structure_oM/Loads/BarVaryingDistributedLoad.cs
```diff
@@ -71,6 +73,10 @@ public class BarVaryingDistributedLoad : BHoMObject, IElementLoad<Bar>
         [Description("If true the load is projected to the element. This means that the load will be reduced when its direction is at an angle to the element.")]
         public virtual bool Projected { get; set; } = false;
 
+        [Description("If true, the StartPosition and EndPosition will be normalised lengths where 0 means start and 1 means end.\n" +
+                     "If false, the StartPosition and EndPosition will be meassured in metres. ")]
+        public virtual bool RelativeDistance { get; set; } = false;
```
peterjamesnugent, line 78: Agree with comment that parametric distances should be the default.

Perhaps rename to `AbsoluteDistance`

I also think we want to avoid the use of units as we have across the rest of the descriptions.

`"If true, the StartPosition and EndPosition will be measured as an absolute distance from the StartNode and EndNode respectively." + 
"If false, the StartPosition and EndPosition will be measured as normalised distances from the StartNode (0) and EndNode (1) respectively."\n"

IsakNaslundBh:
@BHoMBot check installer

peterjamesnugent (approved):
LGTM.

IsakNaslundBh:
@BHoMBot check installer

PR #1093, opened 2020-11-17
Reflection_oM: Add new attributes for supporting fragments

### LINKED ISSUE #478: Reflection_oM: Add Fragment attributes
A series of attributes to be added to be used on `Fragments`. Names below to be confirmed:

- `FragmentDesignation` - Stating which object/interface that this fragment can be applied to
- `UniqueFragment` -Stating that only one of this fragment type can be applied to the specific objects.
- `DefiningFragment` - Stating that this fragment should be seen as something with as high significance as a primary property of the object it is attached to.

Attributes to be used by various engine methods to try to ensure that the fragments are used and assigned correctly

adecler:

Closes #478 

Add the attributes requested in issue #478 

### Test files
Just test that the attributes are having the correct effet in the UI. I''l add screenshots of example of what I tried but you do your own test.


adecler:
Example for Target Attribute:


By adding this attribute in the code:
image

We get:
image

adecler:
Example for unique attribute:

image

Gives us :

image

IsakNaslundBh:
Note on the unique attribute, if replace is true, Would be good to for example if you have some Concrete on a material and assign a Steel with replace set to true, good to replace the Concrete with the Steel and add a note to the user that this has happened.

adecler:
AS discussed earlier, The third attribute (DefiningFragment) is not going to be added (at least in this PR).
Instead I will explore better way the explode component can handle `FragmentSet` (and possibly Fragments themselves).

With that in mind, this PR is ready for review (well, accounting for pending correction pointed out by Isak)

adecler:
@IsakNaslundBh , I have not covered the case when the fragment replaces an existing one. This obviously applies only when the type of the replaced one matches exactly the type of the new one.

adecler:
This PR hasn't been reviewed nor merged in 3 weeks and is now past the deadline for feature PRs on this milestone.
Since it doesn't seem to be as urgent as we initially thought, I will park this for now

IsakNaslundBh:
Took the liberty to update from master on PR merging day, as well as making a simplification of the Constructors for the `Targets` attribute to be using params. This allows for the previous versions to work, but also something like:

image

IsakNaslundBh (approved):
Tested with the engine using 

image

and 

image


and all works as expected (caveat on my approval on the engine).

Approving, but as I made the last commit, good if @FraserGreenroyd could review as well.

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

Reflection_oM/Attributes/TargetsAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+        /***************************************************/
+
+        public virtual List<Type> ValidTypes { get; private set; } = new List<Type>();
+
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public TargetsAttribute(params Type[] types)
```
FraserGreenroyd, line 48:
```suggestion
        public TargetsAttribute(params Type[] validTypes)
```

This ought to resolve the compliance issue as this is an `IImmutable` object

Reflection_oM/Attributes/UniqueAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Reflection.Attributes
+{
+    [AttributeUsage(AttributeTargets.Interface | AttributeTargets.Class)]
+    [Description("Means that there can only be one instance of that type in a collection. This is currently used specifically for the case of fragments.")]
+    public class UniqueAttribute : Attribute, IImmutable
```
FraserGreenroyd, line 35:
```suggestion
    public class UniqueAttribute : Attribute
```

This ought to resolve this one, given the object has no properties that are `get` only.

Now I know the argument back is the object has no properties period, but, if the object has no properties period, then why does it need to be tagged to say that some properties need controlled change (no public `set`)? :wink:

FraserGreenroyd (changes requested):
This looks good and works fine in testing with @IsakNaslundBh , just a couple of suggestions to restore compliance.

If @adecler is happy with the approach taken by @IsakNaslundBh then I'd say this can be merged once compliance is happy, but if not then we'll go back to the discussion board on the implementation :smile:

Reflection_oM/Attributes/TargetsAttribute.cs
adecler: did this in a separate commit as you forgot to also change the name in the body of the constructor

FraserGreenroyd:
@BHoMBot check compliance

Reflection_oM/Attributes/UniqueAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+        /***************************************************/
+
+
+        /***************************************************/
+        /**** Constructors                              ****/
+        /***************************************************/
+
+        public UniqueAttribute()
+        {
+        }
```
FraserGreenroyd, line 48:
```suggestion
```

This file is getting very empty to be fair, but it's for a tag that doesn't need anything so it's still valid.

FraserGreenroyd (changes requested):
Suggested compliance fix missed from last night.

FraserGreenroyd:
image

This still works/looks good following last change.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check core

FraserGreenroyd (approved):
This LGTM now :smile:

PR #1118, opened 2020-12-03
BHoM: IAdapterId and IPersistentId interfaces to require only get

### LINKED ISSUE #1115: BHoM: IPersistentId interface to require only get (not set)
#### Broken rules:

As title, this is required to allow for identifiers that implement also `IImmutable`. The `set` is not required from PersistentId.

As title.

alelom:

Closes #1115 

This is to allow for Identifier objects that implement also `IImmutable`. 
The `set` is not a requirement from the perspective of the `IAdapterId`/`IPersistentId` interfaces.


alelom:
Closed because factually wrong. Contradicting myself.

PR #1119, opened 2020-12-03
BHoM: IPersistentId interface to require only get;

### LINKED ISSUE #1115: BHoM: IPersistentId interface to require only get (not set)

alelom:

Closes #1115 

Required for some Id objects that also implement IImmutable.
See https://github.com/BHoM/Revit_Toolkit/pull/922.


alelom:
Parked for now.

PR #1124, opened 2020-12-11
BHoM: reconsidering representation

rolyhudson:
Prototype for discussion for issue https://github.com/BHoM/BHoM_Engine/issues/2234
See: https://github.com/BHoM/BHoM_Engine/pull/2235
And: https://github.com/BHoM/Grasshopper_Toolkit/pull/577


Graphics_oM/GeometricRepresentation.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace BH.oM.Graphics
+{
+    public class GeometricRepresentation : IRepresentation
+    {
+        public virtual IGeometry Geometry { get; set; } = null;
+
+        public virtual Color Colour { get; set; } = Color.FromArgb(80, 255, 41, 105);
+
+        public virtual int LineWeight { get; set; } = 2;
+    }
```
alelom, line 18: Might need separated settings for contours and surfaces.
E.g. 
LineColour
SurfaceColour

Graphics_oM/IRepresentation.cs
```diff
@@ -0,0 +1,15 @@
...
+using BH.oM.Geometry;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Graphics
+{
+    public interface IRepresentation : IObject
```
alelom, line 11: I'd place this into the base BHoM instead.

Graphics_oM/FontConfig.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Graphics
+{
+    public class FontConfig : IObject
```
alelom, line 11: A couple of comments:
1. I see these properties more as properties of a text rather than of a Font; so in case I'd call this `TextConfig`, if anything
2. Since these are actually defining the properties of a Text on screen, I would probably prefer to see them together with the `TextRepresentation` object rather than in a separate class. 
3. I think I see `...Config` classes more for representing configurations destined for consumption from an engine or adapter method, rather than for composing objects (adding additional properties to another one). If seen in this light, this one here feels rather more like a Fragment.

Graphics_oM/TextRepresentation.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.oM.Graphics
+{
+    public class TextRepresentation : IRepresentation
+    {
+        public virtual string Text { get; set; } = "";
+
+        public virtual Cartesian Cartesian { get; set; } = new Cartesian();
```
alelom, line 15: I'm not sure a cartesian position should be a defining property of a TextRepresentation object. Maybe we could add its positioning as a Fragment? I can imagine different positioning systems may be needed depending on the usage.

alelom (commented):
Thanks for this! I left some comments, happy to chat if you want 😄

rolyhudson:
Closing this as it has been implemented in https://github.com/BHoM/BHoM/pull/1191

PR #1154, opened 2021-01-14
Test_oM: Implement IEvent interface

### LINKED ISSUE #1153: Test_oM: IEvent interface
Workshop with @adecler @IsakNaslundBh @rolyhudson @pawelbaran @LMarkowski @alelom @al-fisher 

A test result should contain a list of `IEvent` that allows individual disciplines to implement their own event for a test result

IEvent enforces these properties:

 - Message
 - Event Type (error? warning? additional types?)
 - Time of test (UTC only)

Then individual event objects can include additional data specific to that event.

`Result` objects then contain properties:

 - Result of test (pass, warning, fail)
 - Description of test

FraserGreenroyd:
Fixes #1153 

### Additional comments

@adecler please check this matches your thoughts from the discussion (the discussion carried on for another half hour after you left but I feel this is in line with what you were thinking).

Versioning - because this is in the 4.0 beta, versioning may be required. However, we (@IsakNaslundBh and I) are unsure of whether we we will change these objects further. So the plan is to handle this PR, close out #1153, and continue with the plan during this milestone, and then add versioning towards the end closer to the 4.1 beta release. This may cause issues @adecler if you have serialised any test results previously - if so then we can handle versioning at the same time as this PR.

Once this PR is merged, @IsakNaslundBh will then update his test objects, and I will update Test_Toolkit for compliance results, so it would be good if this PR didn't hang for too long :wink:


FraserGreenroyd:
@BHoMBot check compliance

adecler (commented):
Ideally, I would like the reflection Event to be an IEvent to avoid having to convert them. That means:
- IEvent is in the core oM
- `Time` should be called `UtcTime` to avoid confusion
- `EventStatus` should contain `Note` and ideally `Error` instead of `Fail` for better backward compatibility. 
Also keep in mind that the .ci solutions will have to be updated otherwise they will no compile. Happy to do it of course. Just waiting on a final agreement on this before making the changes though.

FraserGreenroyd:
> * IEvent is in the core oM

Done

> * `Time` should be called `UtcTime` to avoid confusion (`Event` already has a `Time` and a `UtcTime` properties)

Done, but have called it `UTCTime` owing to `UTC` being an acronym

> * `EventStatus` should contain `Note` and ideally `Error` instead of `Fail` for better backward compatibility.

Done

> Also keep in mind that the .ci solutions will have to be updated otherwise they will no compile. Happy to do it of course. Just waiting on a final agreement on this before making the changes though.

That is expected and captured in the comment on #1153 that we'll be updating a few parts of the code around this, which is why I'd like us to be happy with this and merge this before we update the .ci tests, unit tests, etc. :smile:

> Finally, I am not too sure we need both Message and Description in `TestResult`

I have updated the `Description` attribute for the `Description` to better describe what is expected from the `Description` property - namely the difference is that `Message` (which comes from `IEvent`) is for why the result is `Error` or `Warning` and should be informative of how to change the result, while `Description` describes the test (because we don't have a parent `Test` object that owns a `TestResult` because it doesn't seem necessary to have one) for a CI process.

Hope that helps answer the comments and the changes to the code reflect what you expected @adecler :smile:

FraserGreenroyd:
@BHoMBot check compliance

Test_oM/Results/TestResult.cs
```diff
@@ -30,34 +30,46 @@
...
         public virtual string Description { get; } = "";
 
         [Description("States whether the test was a success or not.")]
-        public virtual ResultStatus Status { get; } = ResultStatus.Undefined;
+        public virtual EventStatus Status { get; set; } = EventStatus.Undefined;
 
         [Description("Events generated during the test.")]
-        public virtual List<Event> Events { get; }
+        public virtual List<IEvent> Events { get; } = new List<IEvent>();
+
```
IsakNaslundBh, line 51: Can we add a `Id` property to this class as well please? String will work fine for what I need it for

FraserGreenroyd: No but we can have an ID property :wink:

adecler:
>>  - Time should be called `UtcTime` to avoid confusion (`Event` already has a `Time` and a `UtcTime` properties)
>
> Done, but have called it `UTCTime` owing to UTC being an acronym

Well, it doesn't matter if it's a completely different name or just a different casing, The Existing `Event` class can only implement the `IEvent` interface if it is exactly the same name. So calling it `UTCTime` doesn't help me at all 😄

FraserGreenroyd:
> > > * Time should be called `UtcTime` to avoid confusion (`Event` already has a `Time` and a `UtcTime` properties)
> > 
> > 
> > Done, but have called it `UTCTime` owing to UTC being an acronym
> 
> Well, it doesn't matter if it's a completely different name or just a different casing, The Existing `Event` class can only implement the `IEvent` interface if it is exactly the same name. So calling it `UTCTime` doesn't help me at all 😄

Early morningness I guess 🤕 

Though I don't like `UtcTime` due to the acronym, I have changed it for now - we can worry about `Utc` vs `UTC` later :smile:

IsakNaslundBh:
> Ideally, I would like the reflection Event to be an IEvent to avoid having to convert them.

Will it not be confusing to have "Pass" as a potential option for the reflection events? Like what will that mean for the UI in terms of messaging?

Same goes the other way around. "Note" does not feel like it belongs as well as a test result? And Error vs Fail is not as distinct IMO.

To me it kind of feels cleaner to keep them separate and to just have a simple method to convert from a debug logging event to a test result event.

But that is just my two pence. If you disagree, can live with it going ahead as is.

FraserGreenroyd:
> Will it not be confusing to have "Pass" as a potential option for the reflection events? Like what will that mean for the UI in terms of messaging?
> 
> Same goes the other way around. "Note" does not feel like it belongs as well as a test result? And Error vs Fail is not as distinct IMO.
> 
> To me it kind of feels cleaner to keep them separate and to just have a simple method to convert from a debug logging event to a test result event.
> 
> But that is just my two pence. If you disagree, can live with it going ahead as is.

Following some of the discussions today with you @IsakNaslundBh and the refactoring of BHoMBot that I've been doing, I'm inclined to agree.

The Reflection_oM Event does seem more suited for its original purpose - a code based event during code operation, as opposed to the Test_oM (Base_oM) Event which is more for the purposes of measuring code.

I would agree, `note` does not make sense for a `Test Result Event` but does for a code operation event.

Maybe it's the naming which leads it to suggest that a debug `Event` could be an `IEvent` - if we renamed `IEvent` -> `ITestEvent` it would be a more clearer separation between the two (and moving it back to Test_oM as a result). There are times I think when a debug event is needed as a test event, and maybe a conversion method would work better than attempting to shoehorn `IEvent` onto `Event.cs`.

There would need to be refactor anyway, as `EventType` cannot be the called that at the moment as `BHoM_UI` (and likely others) get confused between `Base.EventType` and `Reflection.EventType`, hence why `EventType` is currently `EventStatus` (which from a wording perspective makes more sense anyhow to me, a Test Result Event has a status because it's fixed at the time it was generated. I can't find an argument for why Type makes sense though other than it is classifying the debug event, but it could be equally a debug event status. But that is why I actually care less about `EventType` vs `EventStatus` on naming - 6 of 1, half a dozen of the other but our vocabulary ought to be as correct as possible as we roll out testing further this milestone.

But if `IEvent` goes on `Event` it ends up having to implement `EventStatus` which gives a debug event a Status and a Type - which does seem odd (especially given the options are currently matching as well).

So either we refactor the debug Event object in this PR (and then everywhere else it's needed) in one go, or we reconsider if a reflection event ought to be an `IEvent` at all (or maybe a third option, to avoid a false dichotomy).

I dunno if any of that makes sense, it's been a long day with BHoMBot and sprint coordination meetings! Might be one to discuss verbally if it doesn't make sense 😖

FraserGreenroyd:
New thought had this morning, based on the lines that `IEvent` might not be the right name for a Test Event because an `Event` can be almost anything which occurs that we might want to track/log.

For example, BHoMBot wants to track events it fires during its operations (when something was requested, when it was actioned, when it finished, etc.) which are Events, but not test events.

Same with the Reflection Event being a logging event of something occurring during operation rather than testing (but that on occasions it can be used to inform a testing event).

But not all Events, particularly those run from BHoMBot need or care for a status for example.

So I'm wondering if a better approach would be to have an `IEvent` in Base_oM that is much smaller in terms of required fields, that allows us to tag all event based objects (which might work with @alelom event driven stuff he discussed last week), with an `ITestEvent` that implements `IEvent` but requires more specifics for specific events related to testing.

Then Reflection_oM Event could be an `IEvent` (though still would not be suitable directly for testing, but a conversion method can be made), @alelom can build up his event driven stuff, the testing side can build up their test events, without attempting to potentially shoehorn one into the other?

What do people think on that @adecler @IsakNaslundBh ?

FraserGreenroyd:
Following call discussion with @adecler and @IsakNaslundBh :

 - Implement `ITestInformation` in Test_oM - remove `IEvent` from Base_oM
 - `ITestInformation` contains current `IEvent` stuff
 - `Note` == `Pass`
 - `EventStatus` to become `TestInformationStatus` with `Pass`, `Warning`, `Fail` (no `Undefined` option) - default is `Error`

I will get this done today for review ASAP :smile:

FraserGreenroyd:
@adecler @IsakNaslundBh updates made following our call - take a look when you get chance (though ASAP please :wink: )

Descriptions on `TestInformationStatus` - help and suggestions welcome on that!

FraserGreenroyd:
@BHoMBot check compliance

Test_oM/TestInformationStatus.cs
```diff
@@ -0,0 +1,43 @@
...
+using BH.oM.Base;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Linq;
+
+namespace BH.oM.Test
+{
+    [Description("Defines the status of the Test Information. Designed to be very clear, concrete statuses, with minimal ambiguity.")]
+    public enum TestInformationStatus
```
IsakNaslundBh, line 32: Would not `TestStatus` be enough for this?

Test_oM/Results/TestResult.cs
```diff
@@ -29,35 +29,48 @@
...
         public virtual string Description { get; } = "";
 
-        [Description("States whether the test was a success or not.")]
-        public virtual ResultStatus Status { get; } = ResultStatus.Undefined;
+        [Description("States whether the test was a success or not. Defaults to Error to ensure any defaults are investigated appropriately.")]
+        public virtual TestInformationStatus Status { get; set; } = TestInformationStatus.Error;
 
         [Description("Events generated during the test.")]
-        public virtual List<Event> Events { get; }
+        public virtual List<ITestInformation> Events { get; } = new List<ITestInformation>();
```
IsakNaslundBh, line 48: Wonder if we should change the name of this, now that it is no longer a list of `Event`/`IEvent`.

Again, struggling to come up with a better name though, but maybe just `Information` ?

Test_oM/ITestInformation.cs
```diff
@@ -20,22 +20,26 @@
...
-        Fail,
-        CriticalFail,
-    }
-}
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        string Message { get; }
+        TestInformationStatus Status { get; set; }
```
IsakNaslundBh, line 39: Do not think this should be set-able on the level on the interface?

If some class that implements this wants to add a setter, that is ofc fine, but on the base interface I would keep this get only, especially as it is forcing a setter onto the `TestResult` if we have a setter here.

```suggestion
        TestInformationStatus Status { get; }
```

IsakNaslundBh (changes requested):
Generally like it!

Some comments below.

Test_oM/ITestInformation.cs
FraserGreenroyd: Because we are defaulting to an `Error` status I opted to make it easy for people to change this status as this might change far more easily as a result of examination than the other properties.

Test_oM/TestInformationStatus.cs
FraserGreenroyd: We did discuss it in the call, the verbose name seemed to be more desired, especially as this is not the status of a test, but the status of test information. Happy to change to `TestStatus` though if you want.

Test_oM/Results/TestResult.cs
FraserGreenroyd: Agree `Information` doesn't sound 100% but is better than `Events`. Have changed.

Test_oM/TestInformationStatus.cs
FraserGreenroyd: Opted to make the change after all.

Test_oM/ITestInformation.cs
IsakNaslundBh: For some implementation for some particular class that implements this interface, I get your point, but as I stated above, there is nothing stopping an implementer to add a setter for that case.

This is forcing you to add one though, and making this settable almost to me makes the point of making the `TestResult` class immutable mute, as this fundamentally is the result itself. To me it feels like saying that the absolute key bit of information is adjustable, while the adjoining bit is not, which I am not sure I really agree with.

FraserGreenroyd: Ah ok I misunderstood, I thought you were advocating for removing it completely. See latest comment, removed `set` from interface.

Test_oM/Results/TestResult.cs
```diff
@@ -29,35 +29,48 @@
...
         /***************************************************/
 
-        [Description("Describes the focus of the test.")]
+        [Description("Describes the focus of the test that this Test Result answers. This is typically for CI processes but should be human readable for ease of use.")]
         public virtual string Description { get; } = "";
 
-        [Description("States whether the test was a success or not.")]
-        public virtual ResultStatus Status { get; } = ResultStatus.Undefined;
+        [Description("States whether the test was a success or not. Defaults to Error to ensure any defaults are investigated appropriately.")]
+        public virtual TestStatus Status { get; set; } = TestStatus.Error;
```
IsakNaslundBh, line 45: > Ah ok I misunderstood, I thought you were advocating for removing it completely. See latest comment, removed set from interface.

For this class I actually am, reasons being:

- This is the key piece of information. It should not be tempered with
- It is the first argument of the constructor that you _have_ to use. You thereby have to set this value explicitly when you create the result, there is no way to get any default value assigned.
- You really should be knowing if a test has passed or not when you generate this class, otherwise, what is the point..?

FraserGreenroyd: Hmm I'll be honest I don't overly want `TestResult` to be `IImmutable`, I only kept it because it was there originally. `IImmutable` only brings benefits to a UI whereas this is almost exclusively going to be a code based object (though not impossible to do within the UI if really wanted. I'd much rather be able to provide `set` to all these properties (but I'm biased, it makes my life easier from the BHoMBot/coding perspective).

I would argue there may be acceptable reasons to not yet know the result of the class at the time its generated, particularly in the workflows that BHoMBot goes through where not all of the information is known until the end and a final decision has to be made on what's available, but an object to track a test/result needs to go through the system.

IsakNaslundBh: > Hmm I'll be honest I don't overly want TestResult to be IImmutable

Could potentially get behind it not being `IImmutable`. Very much prefer that to the current version with some things settable, some things not. To me this is quite a all or nothing situation. To me, the latest change (making the information settable) puts it even more in a weird in-between state. 

Would say, either we make all settable and remove the `IImmutable` interface, or we remove all setters.

@adecler any preference?

FraserGreenroyd: I agree, although there are plenty of `IImmutable` objects which have set accessors (rightly or wrongly) within other parts of BHoM. To come back to the use of `IImmutable` from the implementations, is to prevent dependant properties being set independently of each other, such as G-Value and U-Value being directly linked, updating one risks the other being out of sync, so the object which has them is `IImmutable` to prevent that.

But in this case, we run a test, we get information that then determines the result of the test, the evaluation of the information is important and might not be able to be done until the end. Take [this PR from this morning](https://github.com/BHoM/BHoM/pull/1161/files) as an example. On the files changed you can see BHoMBot has run a test and reported an error. But more information is given (in this case, a human reviewer providing an override) that changes the result of the test but not the rest of the information. The test hasn't changed, but the information has and as a result, the status has with it.

IsakNaslundBh: Yeah, if we need it not to be immutable, then I am fine with that. It was the mix that I had issues with, especially as the mix did exclude the most critical part. 😄 

If @adecler is happy with this not being immutable, then I am happy with this as well.

adecler: Very happy with it not being `IImmutable` as I don't see any reason for it to be. 
As far as I am concerned, the only case where we should have not-settable properties (an therefore `IImmutable` object) is when those properties are not orthogonal to each other. Anything else is just purely founded on this weird idea that you need to take away agency from the users for "their own sake" (and the guilty pleasure of pissing them off)

FraserGreenroyd: Ok great, sounds like we're all on the same page then :smile: any chance one of you could review and either request more changes or approve so we can close this out? 👼

Test_oM/Results/TestResult.cs
```diff
@@ -29,38 +29,33 @@
...
-        public virtual string Description { get; } = "";
+        [Description("Describes the focus of the test that this Test Result answers. This is typically for CI processes but should be human readable for ease of use.")]
+        public virtual string Description { get; set; } = "";
 
-        [Description("States whether the test was a success or not.")]
-        public virtual ResultStatus Status { get; } = ResultStatus.Undefined;
+        [Description("States whether the test was a success or not. Defaults to Error to ensure any defaults are investigated appropriately.")]
+        public virtual TestStatus Status { get; set; } = TestStatus.Error;
 
         [Description("Events generated during the test.")]
```
adecler, line 47: Should not be called `Events` in the description anymore

adecler:
Looks good to me from visual inspection. 
Will do a proper review tomorrow but happy for you to merge if @IsakNaslundBh approves the PR

Test_oM/Results/TestResult.cs
IsakNaslundBh: Yeah, happy as well!

IsakNaslundBh (approved):
This looks good to me now. Think what this allows should be flexible enough to work with what I have done for the push-pull-compare as well as the UnitTests!

FraserGreenroyd:
@BHoMBot check compliance

PR #1161, opened 2021-01-20
Reflection_oM: Fix non compliant IImmutable attributes

### LINKED ISSUE #1160: Reflection_oM: InputAttribute, OutputAttribute and MultiOutputAttribte are not complaint IImmutables
#### Broken rules:
As per title, they are missing a valid constructor mirroring the properties

Add the missing constructor

adecler:

Closes #1160

After the change on the Quantity property of those attributes, the classes became non compliant (missing matching constructor) and therefore failed to deserialise correctly.

### Additional comments
Note that, to deserialise correctly, we also need a correction of the property name in the versionning `Upgrade.json` file. Will push soon with a series of other fixes on versioning once the current PR is merged

### DIFF
--- a/Reflection_oM/Attributes/InputAttribute.cs
+++ b/Reflection_oM/Attributes/InputAttribute.cs
@@ -56,6 +56,15 @@ namespace BH.oM.Reflection.Attributes
 
         /***************************************************/
 
+        public InputAttribute(string name, string description, InputClassificationAttribute classification, Type typeId)
+        {
+            Name = name;
+            Description = description;
+            Classification = classification;
+        }
+
+        /***************************************************/
+
         public InputAttribute(string name, string description)
         {
             Name = name;
--- a/Reflection_oM/Attributes/MultiOutputAttribute.cs
+++ b/Reflection_oM/Attributes/MultiOutputAttribute.cs
@@ -58,6 +58,16 @@ namespace BH.oM.Reflection.Attributes
         }
 
         /***************************************************/
+
+        public MultiOutputAttribute(int index, string name, string description, InputClassificationAttribute classification, Type typeId)
+        {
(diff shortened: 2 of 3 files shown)


FraserGreenroyd:
@BHoMBot check code compliance

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 1733642681

Reflection_oM/Attributes/InputAttribute.cs
```diff
@@ -56,6 +56,15 @@ public InputAttribute(string name, string description, Type classification = nul
 
         /***************************************************/
 
+        public InputAttribute(string name, string description, InputClassificationAttribute classification, Type typeId)
```
FraserGreenroyd, line 59: We don't have a `Type` property and `typeId` isn't being used in the constructor, is this just for the versioning? Seems to be the same on the others too?

FraserGreenroyd (commented):
Question raised, am happy to approve as is but just want to double check on the question :smile:

Reflection_oM/Attributes/InputAttribute.cs
adecler: Yep, it's just for the serialisation to work. Tried quite a few alternatives and this was the only one that was working sadly.

adecler: `typeId` property is in the parent class

FraserGreenroyd (approved):
LGTM based on discussion :smile:

FraserGreenroyd:
@BHoMBot check copyright compliance

FraserGreenroyd:
@BHoMBot check documentation compliance

PR #1165, opened 2021-01-21
Data_oM: Added "domain extremes" property to Domain class

### LINKED ISSUE #1162: Data_oM: Domain object does not define whether extremes are included or not
#### Description:

As title. This is a problem because:
- certain checks may require extremes to be included or not included
- it is not clear what the assumption is - are extremes included or not by default?

## 1) Default value

Currently, methods in the BHoM_Engine like `IsInRange` seem to default to the fact that the Domain does **not** include the extremes.

This may be seen as counter-intuitive e.g. Grasshopper does include the extremes in the domain:
image

## 2) Make the "extremes inclusion" a defining property of the Domain 
Suggestion to add a `DomainExtremesInclusion` enum property to the Domain:

```cs
    public enum DomainExtremesInclusion
    {
        IncludeBoth,
        IncludeMin,
        IncludeMax,
        ExcludeBoth
    }
```

this would remove any confusion around the extremes and give the option on whether to include them or not, and which.

alelom:

Closes #1162

### Test files

See https://github.com/BHoM/BHoM_Engine/pull/2283


alelom:
Parked for now.

PR #1173, opened 2021-01-28
Programming_oM: Remove ParentId property from INodeParam

### LINKED ISSUE #1172: Programming_oM: Remove ParentId from INodeParam classes
#### Broken rules:
The `ParentId` property creates this awkward situation where the immutability needs to be enforced across multiple objects.
This scenario is not supported in the BHoM and is actually not necessary in this case.

`ParentId` is used in a single method inside the CSharp toolkit. That method can actually function without the `ParentId` property (by storing that information elsewhere). Fox that method and then remove every reference to `ParentId`

adecler:
### NOTE: Related PRs 

https://github.com/BHoM/BHoM_Engine/pull/2286
https://github.com/BHoM/CSharp_Toolkit/pull/41
https://github.com/BHoM/Grasshopper_Toolkit/pull/592
https://github.com/BHoM/Versioning_Toolkit/pull/135
   
Closes #1172 

See issue. Pretty straightforward removal of a property. The key thing to test is that the Node2Code component is still working correctly as this is the only place where `ParentId` was actually used.

@rolyhudson , this should probably make the most sense to you as this is very similar to what we did on the GraphFlow stuff.

@FraserGreenroyd , @IsakNaslundBh , adding you more as a FYI as the whole concept of Immutability is close to you. This was an interesting case as there wasn't really a good combination of  constructor and get only properties that would prevent all possible invalid parentId values. The CreateDummy` method from the Test toolkit was the first to highlight that.

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/Installers/Scripts/BuroHappold_BHoM_v3.3/0066_CSharp_Toolkit/CSharp_Toolkit.gh?csf=1&web=1&e=Iur2MX

### DIFF
--- a/Programming_oM/Nodes/BlockNode.cs
+++ b/Programming_oM/Nodes/BlockNode.cs
@@ -33,13 +33,13 @@ using System.Threading.Tasks;
 namespace BH.oM.Programming
 {
     [Description("Represents a group of syntax nodes covered by a common description. This is equivalent to a block of code inside a method.")]
-    public class BlockNode : BHoMObject, INode, IImmutable
+    public class BlockNode : BHoMObject, INode
     {
         /***************************************************/
         /**** Properties                                ****/
         /***************************************************/
 
-        public virtual List<INode> InternalNodes { get; } = new List<INode>();
+        public virtual List<INode> InternalNodes { get; set; } = new List<INode>();
 
         public virtual string Description { get; set; } = "";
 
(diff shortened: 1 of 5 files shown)


IsakNaslundBh:
> @FraserGreenroyd , @IsakNaslundBh , adding you more as a FYI as the whole concept of Immutability is close to you. This was an interesting case as there wasn't really a good combination of constructor and get only properties that would prevent all possible invalid parentId values. The CreateDummy` method from the Test toolkit was the first to highlight that.

Thanks @adecler . On the level of the constructor, I do not think we really can protect against invalid objects though. The `CreateDummy` method is creating invalid SteelSections as well (A circle with a diameter of 42.42m does not have an area of 42.42 m2 etc), and same can easily be done from the side of the code. I have instead always looked at it being the `Create` methods responsibility to make sure they output something valid, as they can do all the checking etc.

This might be beyond the point of this PR though, but just to answer your comment.

adecler:
@IsakNaslundBh , what I meant was that the BHoMGuid of the node as well as the ParentId of a INodeParam was settable. This means that the object can be rendered invalid after construction by setting one of those properties. 

Immutable objects don't have their constructors directly exposed to the UI so the Create methods are in charge of preventing the creation of invalid objects.

CreateDummy is obviously a bit of a weird case because it has no choice but to access those constructors directly (well, I guess, it could look for a Create method in the engine) but the part that was relevant to me is that it manually sets all the settable properties AFTER creating the object from the installer.  I don't remember why that decision was made, but it is certainly great for debugging !

rolyhudson (approved):
Ran test files and found behaviour as expected. 
Reviewed code - logic change to storing of `ParentId` on the `ReceiverState` makes sense.
Personal appreciation of Cluster2Code nudged up to next level.

FraserGreenroyd (approved):
Code review as no test files provided, but LGTM and works with other PRs in the series and versioning.

FraserGreenroyd:
@BHoMBot check compliance

adecler:
> Code review as no test files provided, but LGTM and works with other PRs in the series and versioning.

image

PR #1186, opened 2021-02-02
Architecture_oM: BuildersWorkOpening class added together with necessary references

### LINKED ISSUE #1185: Architecture_oM: Create BuildersWorkOpening object
#### Description:

After a discussion with @IsakNaslundBh, I came up with a proposal of a class `BuildersWorkOpening` with following properties:
- `BH.oM.Spatial.ShapeProfiles.IProfile Profile` 
- `double Depth`
- `BH.oM.Geometry.CoordinateSystem.Cartesian CoordinateSystem`

Although the object was originally considered to be placed in Physical_oM, it does not fully fit to that project (the opening is not always physical + it could cause confusion with `IOpenings`). @al-fisher FYI

pawelbaran:

Closes #1185

### Changelog

- `Opening` class added to Architecture_oM

### Additional comments

Although the originally discussed location was Physical_oM, together with @IsakNaslundBh we concluded it does not fully suit the namespace because:
- it may not always be physical (just provisional)
- it may cause confusion with `IOpening` 
- can go out of sync in case of being hosted on a physical wall

I picked Architecture_oM as the target project as the problem of builders work is transdisciplinary, so that was my best guess. However, happy to move it over to other location (maybe Coordination_oM?) if requested by the reviewers (@al-fisher 🙈).


Architecture_oM/BuildersWork/BuildersWorkOpening.cs
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System.ComponentModel;
+
+namespace BH.oM.Architecture.BuildersWork
+{
+    [Description("Object representing a builders work opening as an entity independent of its host.")]
+    public class BuildersWorkOpening : BHoMObject
```
FraserGreenroyd, line 30: If we have to use the term `BuildersWork` then I wouldn't want the object to also include it when the namespace is. Otherwise the full definition is `BH.oM.Architecture.BuildersWork.BuildersWorkOpening`. Context from the namespace says it will be a `BuildersWork` object.

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /***************************************************/
+        /****                Properties                 ****/
+        /***************************************************/
+
+        [Description("Profile of the opening.")]
+        public virtual BH.oM.Spatial.ShapeProfiles.IProfile Profile { get; set; } = null;
```
FraserGreenroyd, line 37: Why not `using BH.oM.Spatial.ShapeProfiles` at the top to make this tidier?

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        [Length]
+        [Description("Total depth of the opening, distributed equally from the origin in both directions along the local z axis.")]
+        public virtual double Depth { get; set; } = double.NaN;
+
+        [Description("Local coordinate system of the opening with local z being the direction along the opening, local x representing direction along width in plane and y along height respectively.")]
+        public virtual BH.oM.Geometry.CoordinateSystem.Cartesian CoordinateSystem { get; set; } = new Geometry.CoordinateSystem.Cartesian();
```
FraserGreenroyd, line 44: Same as above?

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
```diff
@@ -0,0 +1,50 @@
...
+        /***************************************************/
+        /****                Properties                 ****/
+        /***************************************************/
+
+        [Description("Profile of the opening.")]
+        public virtual BH.oM.Spatial.ShapeProfiles.IProfile Profile { get; set; } = null;
+
+        [Length]
+        [Description("Total depth of the opening, distributed equally from the origin in both directions along the local z axis.")]
+        public virtual double Depth { get; set; } = double.NaN;
```
FraserGreenroyd, line 41: If the depth doesn't exist, is it not a 0 length? Easier to handle in the rest of the code without needing `NaN` checks everywhere where this property would be used? @michaelhoehn knows the pain of that :smile:

FraserGreenroyd (changes requested):
I'm not convinced on the use of `BuildersWork` in the namespace but am happy to let it go for now if the consensus is to use it.

> Test files
> Not needed.

As I said to @alelom who used that line, I disagree :wink:

What is needed, I think, and is missing, is an example workflow that highlights why this object is required. I couldn't see it on the issue and can't see it on SharePoint. If we merge this object in here without knowing how we plan to use it, or seeing how we plan to use it, we could end up with an object we have to maintain that doesn't bring much benefit. Especially given the object would need to be versioned appropriately for any future changes moving forward.

FraserGreenroyd:
@BHoMBot check compliance

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
pawelbaran: zero difference, but made you happy 👍

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
pawelbaran: fair point - fixed

Architecture_oM/BuildersWork/BuildersWorkOpening.cs
pawelbaran: I have changed it to BH.oM.Architecture.BuildersWork.Opening on the branch, but not 100% sure about it - will not be too ambiguous with all other openings we have in the oM?

pawelbaran:
> > Test files
> > Not needed.
> 
> As I said to @alelom who used that line, I disagree 😉

I have adjusted this section of PR description to the standard set by your last BHoM PRs, I hope you will like it more 😃 

> 
> What is needed, I think, and is missing, is an example workflow that highlights why this object is required. I couldn't see it on the issue and can't see it on SharePoint. If we merge this object in here without knowing how we plan to use it, or seeing how we plan to use it, we could end up with an object we have to maintain that doesn't bring much benefit. Especially given the object would need to be versioned appropriately for any future changes moving forward.

The plan is to use this object in the workflows dealing with builders work:
- builders work generation based on MEP/STR collisions
- distance queries between the provisions (fire compliance checks)
- sizing checks etc.

It has been mentioned a few times recently, so you should not be surprised @FraserGreenroyd 🙊

FraserGreenroyd (approved):
> I have adjusted this section of PR description to the standard set by your last 10 BHoM PRs, I hope you will like it more 😃

Touche, but I would have thought you'd want to do better than me @pawelbaran :wink:

> The plan is to use this object in the workflows dealing with builders work:
> 
> * builders work generation based on MEP/STR collisions
> * distance queries between the provisions (fire compliance checks)
> * sizing checks etc.

Ok that makes sense, and following on from our offline conversation this seems appropriate for now and you've accepted the risk of having an object here than prototyped elsewhere.

> It has been mentioned a few times recently, so you should not be surprised @FraserGreenroyd 🙊

I'm not surprised, but someone else might be given the discussion/mentions happened outside the GitHub forums, and it's important we make sure we document those conversations to ensure no one else is surprise :wink:

Otherwise, as discussed, this looks fine to me, will let @al-fisher have final review in terms of whether Architecture is the correct namespace.

al-fisher:
> I picked Architecture_oM as the target project as the problem of builders work is transdisciplinary, so that was my best guess. However, happy to move it over to other location (maybe Coordination_oM?) if requested by the reviewers (@al-fisher 🙈).

Fair enough - certainly makes sense for getting going! Thanks @pawelbaran

al-fisher (dismissed):

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check compliance

al-fisher (approved):
Reapproving after compliance commit

PR #1192, opened 2021-02-10
Physical_oM: Adding reinforcement fragments

### LINKED ISSUE #757: Structure_oM: Add RebarSet object
#### Description:

For handling bundles of Rebars, after discussions with @pawelbaran and @LMarkowski 

proposed class:

```
    public class RebarSet : BHoMObject
    {

        public List<Rebar> Rebars { get; set; }

    }
```

LMarkowski:

Closes #757

<S>This is a draft PR depicting the general idea for the family of RebarSet objects.</S>
Implementing the object model for rebar sets, using fragments, as discussed on last week's chat.
Comments on naming, descriptions etc will be very welecome.
Generally see [comment below](https://github.com/BHoM/BHoM/pull/1192#issuecomment-781964855).
I thought of implementing the engine side alongside with this PR but decided to do it separately as it might be quite a lot of work to do.

### Test files

Will be created during implementation of the engine side.

### Changelog

- adding `IRebarSet` and `IParametricRebarSet` interfaces to the `Physical_oM`.
- adding `ReinforcementFragment`, `ExplicitRebarSet`, `LinearRebarSetAlong` and `LinearRebarSetAcross` classes to the `Physical_oM`.
- adding `ReinforcementFuncion` enumeration to the `Physical_oM`.

### Additional comments

Next steps will be rising (and resolving 😉) issues for:
- Implementing the engine side of the rebar sets.
- Adding parametric sets for reinforcement of 2D elements.
- Exposing shape codes in the oM.


pawelbaran:
I like the concept in general (I co-authored it in the end 😃), one question that I have in mind now is whether it would be OK to use `IReinforcementFragments` as standalone objects, not hosted by any other BHoMObject, in case we wanted to e.g. push the rebar to some software without pairing it up with any elements beforehand? I guess that should be fine code wise (if the fragment inherits from `BHoMObject`), but what about the philosophy?

What are your feelings @al-fisher @IsakNaslundBh @rwemay? I agree with @LMarkowski the chat might be really helpful here.

al-fisher:
Yes - pushing fragments independently is fine. Fragments (as a concept) can happily exist and be pushed etc. independent of an object owning them in their fragment list.
We have examples of Materials, where they are Fragments so _can_ be attached, but can also exist independent.

To support that further I would not suffix all the objects with the word "Fragment". 
They can implement the `IFragment` interface - but would just call them `IReinforcement`, `LinearReinforcement` etc.

LMarkowski:
@IsakNaslundBh @peterjamesnugent @pawelbaran @al-fisher 

So as mentioned on a chat I've created an UML-ish diagram on our ideas for Reinforcement fragments and sets:

reinforcement diagram

Notes:
- `IParametricRebarSet` and its children is almost a copypaste from `Str_oM` `IBarReinforcement`.
- `Linear` and `Surface` names could actually be replaced with `1D` and `2D` regarding the host type (and maybe grouped in separate interfaces?).
- Naming as a whole can be improved by someone better experienced in English and/or engineering 😄 
- `SurfaceRebarSet` is a pure draft as I've done none work on the subject yet, and probably won't be implemented for now. It's just to mark it's place.
- `ExplicitRebarSet` could also be interfaced and split into separate classes, but I feel like marking the function is enough.
- Querying explicit sets from parametric ones should be available. I'm not sure about the other way round. Maybe just for most common cases.
- Next step could be implementing shape codes. It would mean extending `IReinforcingBar` with parametric classes for each shape code and using them instead of (or alongside with) `PrimaryReinforcingBar` and `Stirrup`.

Please shout out if I've missed anything!

Physical_oM/Enums/ReinforcementFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+ */
+ 
+
+namespace BH.oM.Physical.Elements
+{
+    public enum ReinforcementFunction
+    {
+        Undefined,
+        Primary,
+        Stirrups,
```
pawelbaran, line 30: I would call it `Stirrup`

Physical_oM/Enums/ReinforcementFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+ 
+
+namespace BH.oM.Physical.Elements
+{
+    public enum ReinforcementFunction
```
pawelbaran, line 26: It would be nice to have a description for the final version.

pawelbaran (commented):
Looks good in general, I have left two minor code comments. What I am missing (and am most concerned about) is `SurfaceRebarSet`. My concern regarding that type comes from defining it by X and Y parameters - I would consider doing this using a surface, to allow applying it to irregular and non-orthogonal regions.

Physical_oM/Enums/ReinforcementFunction.cs
LMarkowski: Are you sure? I thought if it is a function used in the sets bars would always be multiple.

pawelbaran: Hmm, wouldn't that enum be helpful for single bars?

LMarkowski: It's being used only in sets. I can rename it to `RebarSetFunction` or move it to the `IReinforcingBar` interface.

pawelbaran: At this point I would think of stopping for a while and making sure that the concept is not becoming inconsistent: bar type being driven by type for single items vs enum for rebar sets. I am not necessarily against the current proposal, it is more of a food for thought.

LMarkowski:
Parking this one until there will be enough resources on the structures side to give it a proper review (most probably next milestone).

IsakNaslundBh:
Closing as superseded by work done by https://github.com/BHoM/BHoM/pull/1286

Will keep the branch of this PR open for now, as there might be some bits in here that still could be useful to migrate over.

PR #1229, opened 2021-04-07
Structure_oM: Adding UniformLoadSet and ContourLoadSet from RAM_oM

### LINKED ISSUE #1224: Structure_oM: promote UniformLoadSet from RAM_oM
#### Description:

I've got an object UniformLoadSet:
```
    public class UniformLoadSet : BHoMObject
    {

        public virtual Dictionary<string, Loadcase> Loadcases { get; set; } = null;
        public virtual Dictionary<string, double> Loads { get; set; } = null;

    }
```

which I'd like to use more broadly for load maps in Revit, and eventually for other FE packages which support similar structures (ETABS has Uniform Load Sets but does not expose them in the API in the current version)

JosefTaylor:

Closes #1224

### DIFF
--- a/Structure_oM/Loads/ContourLoadSet.cs
+++ b/Structure_oM/Loads/ContourLoadSet.cs
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2021, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
@@ -44,3 +44,4 @@ namespace BH.oM.Structure.Loads
     }
 
 }
+
--- a/Structure_oM/Loads/ContourLoadSet.cs
+++ b/Structure_oM/Loads/ContourLoadSet.cs
@@ -1,0 +1,46 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Geometry;
+using System.ComponentModel;
+using BH.oM.Quantities.Attributes;
+
+namespace BH.oM.Structure.Loads
+{
+    [Description("A set of uniform loads to be applied in the gravity direction over a region defined by a closed planar Polyline.")]
+    public class ContourLoadSet : BHoMObject
+    {
+        /***************************************************/
+        /****            Public Properties              ****/
+        /***************************************************/
+
+        [Description("A UniformLoadSet containing a list of uniform loads to be applied in the gravity direction")]
+        public virtual UniformLoadSet UniformLoadSet { get; set; } = null;
+
+        [Description("Closed planar polyline enclosing the region to be loaded.")]
+        public virtual Polyline Contour { get; set; } = null;
+
+        /***************************************************/
+    }
+
+}
--- a/Structure_oM/Versioning_42.json
+++ b/Structure_oM/Versioning_42.json
@@ -1,0 +1,24 @@
+﻿{
+  "Namespace": {
+    "ToNew": {
+
+    },
+    "ToOld": {
+
+    }
+  },
+  "Type": {
+    "ToNew": {
+      "BH.oM.Adapters.RAM.ContourLoadSet": "BH.oM.Structure.Loads.ContourLoadSet"
+    },
+    "ToOld": {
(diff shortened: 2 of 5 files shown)


FraserGreenroyd:
@BHoMBot check code-compliance

FraserGreenroyd:
@BHoMBot check code-compliance

JosefTaylor:
@BHoMBot check versioning

Structure_oM/Versioning_42.json
```diff
@@ -0,0 +1,24 @@
...
+    "ToNew": {
+
+    },
+    "ToOld": {
+
+    }
+  },
+  "Type": {
+    "ToNew": {
+      "BH.oM.Adapters.RAM.ContourLoadSet": "BH.oM.Structure.Loads.ContourLoadSet"
```
IsakNaslundBh, line 12: Realised you probably have to add a ToNew and ToOld for `UniformLoadSet` here as well, even though you have a custom upgrader. Reason being that there are actually two things that need to function for the upgrading.

First, that the custom upgrader is handling, is the upgrade of any serialised _objects_ stored somewhere. As actual instances of the object.

Second is the Type, which is _not_ handled by the custom upgrader but it relies only on the system in here. This will be important if you have any method for example relying on the type etc, or any component pointing to the type and so on.

Structure_oM/Loads/ContourLoadSet.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /***************************************************/
+        /****            Public Properties              ****/
+        /***************************************************/
+
+        [Description("A UniformLoadSet containing a list of uniform loads to be applied in the gravity direction")]
+        public virtual UniformLoadSet UniformLoadSet { get; set; } = null;
+
+        [Description("Closed planar polyline enclosing the region to be loaded.")]
+        public virtual Polyline Contour { get; set; } = null;
```
IsakNaslundBh, line 41: Just as a thought, for this to work with filled regions in revit, wonder if this either should be changed to a `PlanarSurface` to allow for openings in the region, or even a `PolySurface` to allow for disjointed regions. The second part could be handled by multiple of this object instead ofc, so not sure if necessary.

This could also be resolved by adding for example a `List<Polyline>` property for openings.

Also, this ofc needs to be thought through in terms of how it will be used when passed to structural packages that might not support holes in the regions as well, so might be better to keep it as is. Mostly raising here as a food for thought if nothing else.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
Also in addition to the compliance that will need resolving at some point before merge, the PR title also needs to reference which oM is being modified :smile:

Structure_oM/Versioning_42.json
FraserGreenroyd, line 12: There needs to be an addition for `BH.oM.Adapters.RAM.UniformLoadSet` added here as well in order for Versioning_Toolkit to detect that change too.

FraserGreenroyd (dismissed):

Structure_oM/Loads/ContourLoadSet.cs
JosefTaylor: agreed, this should map directly to Filled Regions in Revit (though that mapping will generally be custom per script, rather than coded, due to different standards for load maps in Revit).

Structure_oM/Versioning_42.json
JosefTaylor: Thought that was covered in Versioning Toolkit with the upgrader

IsakNaslundBh: @JosefTaylor see my comment here: https://github.com/BHoM/BHoM/pull/1229#discussion_r609420218 :)

JosefTaylor: Fraser noted this as well- I've added the line but it's not seeming to work on that old component - it gives me the magic component error, instead of the MessageForDeleted. I'll resolve this conversation and we can continue it below.

FraserGreenroyd: To be fair to @IsakNaslundBh he did note it first 😋

FraserGreenroyd (commented):
I am acknowledging the request for my re-review following my look at the versioning side, and can confirm that I have looked at this, RAM, and Versioning PRs and am happy with the versioning put in from a CI/CD perspective. I cannot comment on full operation for scripts as that is outside my domain.

For this PR in particular, the compliance will need to pass before merge (notably the copyright headers are displaying the wrong year, BHoMBot can handle this for you if you wish similar to the automatic-full stop adding operation).

Other than that, final review prior to merge will need to be handled by the discipline as this is structures specific work and not generic enough for CI/CD approval alone. As such I am commenting this review as my approval from a CI/CD perspective pending passing required checks, and will leave final review before merge to @IsakNaslundBh and team to avoid an accidental merge without a discipline review. I hope this is ok :smile:

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check core

JosefTaylor:
@BHoMBot fix copyright headers ref. 2301081882

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check core

JosefTaylor:
@BHoMBot check installer

JosefTaylor:
@BHoMBot check serialisation

JosefTaylor:
@BHoMBot check BHoM.CheckInstaller

JosefTaylor:
@FraserGreenroyd @IsakNaslundBh Could you please add your review and merge if satisfied? @maryannewachter has approved the RAM PR including the functionality of the migrated objects.

FraserGreenroyd:
I believe @maryannewachter reported a versioning bug introduced as a result of these PRs? I would not want to see them merged if they are introducing such an issue, but the domain as a whole is structures, so @IsakNaslundBh will have to merge when he's happy. From a CI perspective I'm happy IF there are no versioning issues introduced by these PRs.

JosefTaylor:
@FraserGreenroyd yes, the deleted method doesn't show the MessageForDeleted that I've defined- I'd like to fix that but as far as I can tell I've done it properly in the RAM toolkit. If you can help me out, we can get it fixed for this PR, otherwise I have already raised it as a separate issue in the RAM toolkit.

maryannewachter (approved):
Works per [review](https://github.com/BHoM/RAM_Toolkit/pull/183) of RAM_oM

JosefTaylor:
@BHoMBot check ready to merge

JosefTaylor:
@BHoMBot check ready-to-merge

JosefTaylor:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1247, opened 2021-05-26
Structure_oM: Add ReinforcementDensity, BarRequiredArea and MeshRequiredArea classes

### LINKED ISSUE #1239: Structure_oM: Adding of ReinforcementDensity and RequiredArea classes
#### Description:

First two representations of rebar within the BHoM at the lowest level.

1. Reinforcement density
- Makes use of the existing `Fragments` as it's just storing a single number below
- These are owned by the `Bar` or the `IAreaElement` 
- Being a `Fragment` it can also be mapped to `Physical` elements if required for BIM software

2. Required area
- As discussed, a result class - this allows us to make use of the existing `BarResult` and `MeshResult` classes such as internal divisions (on a `Bar`) and `Node` results on a `Mesh`
image

Open for discussion but I think these two align with what we discussed @pawelbaran @al-fisher @IsakNaslundBh @alelom @rwemay

peterjamesnugent:

Closes #1239 

### Changelog

- Added `ReinforcementDensity` as a `Fragment` to be assigned to `Bar` and `AreaElement` objects
- Added `BarRequiredArea`/`MeshRequiredArea` as a `BarResult`/`MeshElementResult` to receive results from calculations and/or software
- Added `AreaPerUnitLength` attribute to the `Quantities_oM`


peterjamesnugent:
@BHoMBot check compliance

pawelbaran:
@peterjamesnugent, following today's call, I tried to put my thoughts in order, ending up with the table as shown below (source available [on SharePoint](https://burohappold.sharepoint.com/:x:/r/sites/Direction/designtechnology/_layouts/15/Doc.aspx?sourcedoc=%7B32926174-A0F8-43BD-8D04-DC9E686DD72D%7D&file=RepresentationLevels.xlsx&action=default&mobileredirect=true)). Please note that this is just a summary of my ignorant thoughts and is not a change request, rather yet another attempt to map out the problem top-down. Feel free to comment/change/ignore as you definitely know better what would make sense in practice.

image

A few words of explanation:
- red means questions I could not answer myself
- regions == closed `PolyCurves` contained within the outline of the panel (Boolean intersection to be used to check/update)
- as discussed, I started considering 2 levels of required area: one as pure results from FE analysis, the other as their engineering interpretation, i.e. discrete regions with values equal to worst (average?) node values - not sure if this makes sense, maybe I am splitting the hair in four, as we say in Polish 🙈 
- the thick horizontal line means that we have 2 different approaches to representation of reinforcement:
    - preliminary/analytical - high level of abstraction, where each element and its reinforcement can be considered individually and the relationships are managed using releases, supports etc.
    - representation - reinforcement is considered as a system independent of the host elements, e.g. one set of bars can cover 2 neighbouring panels (see top reinforcement in the pic below)
    image
    
        Extra feature of the element-independent rebar representation is the ability to exchange it between the models with mismatching topology (e.g. some panels can be split in FE, but continuous in BIM etc.)

Apologies for the wall of text, but I was really trying to capture all thoughts and put them in the right slots. I will try to join tomorrow's session with @IsakNaslundBh to further explain myself if the vaccine does not kill me - otherwise please plough on without me 👍

peterjamesnugent:
@pawelbaran I've updated the table below based on the discussion with ourselves and @IsakNaslundBh.

image

pawelbaran:
All makes sense, now we only need to implement all that! 😃 Thanks @peterjamesnugent

peterjamesnugent:
@BHoMBot check required

Structure_oM/Results/Bar Results/BarRequiredArea.cs
```diff
@@ -0,0 +1,83 @@
...
+        [Area]
+        [Description("Minimum required area of reinforcement in the bottom layer.")]
+        public virtual double Bottom { get; }
+
+        [Area]
+        [Description("Minimum required area of perimeter reinforcement.")]
+        public virtual double Perimeter { get; }
+
+        [Length]
+        [Description("Minimum required area of shear reinforcement expressed as an area per unit length.")]
```
pawelbaran, line 52: Is mm2/mm same as mm? Theoretically yes, in practice may be misleading. Just flagging up, happy to keep as is if others do not see it as an issue 👍

Structure_oM/Results/Mesh/MeshRequiredArea.cs
```diff
@@ -0,0 +1,92 @@
...
+        [Area]
+        [Description("Minimum required area of reinforcement in the top layer.")]
+        public virtual double Top { get; }
+
+        [Area]
+        [Description("Minimum required area of reinforcement in the bottom layer.")]
+        public virtual double Bottom { get; }
+
+        [Area]
+        [Description("Minimum required area of perimeter reinforcement.")]
```
pawelbaran, line 48: What is perimeter reinforcement here? Call me ignorant, you will be right 🙈

Structure_oM/Results/Mesh/MeshRequiredArea.cs
```diff
@@ -0,0 +1,92 @@
...
+        public virtual double Perimeter { get; }
+
+        [Length]
+        [Description("Minimum required area of shear reinforcement expressed as an area per unit length.")]
+        public virtual double Shear { get; }
+
+        [Area]
+        [Description("Minimum required area of torsion reinforcement.")]
```
pawelbaran, line 56: Same as above, is this a practically used value in wall or slab design?

Structure_oM/Results/Bar Results/BarRequiredArea.cs
```diff
@@ -0,0 +1,83 @@
...
+        public virtual double Shear { get; }
+
+        [Area]
+        [Description("Minimum required area of torsion reinforcement.")]
+        public virtual double Torsion { get; }
+
+        [Description("Material name for the reinforcement.")]
```
pawelbaran, line 59: Why material name and not material as `IMaterialFragment`?

Structure_oM/Results/Mesh/MeshRequiredArea.cs
```diff
@@ -0,0 +1,92 @@
...
+        public virtual double Torsion { get; }
+
+        [Description("Material name for the reinforcement.")]
+        public virtual string MaterialName { get; }
```
pawelbaran, line 60: Why material name and not material as `IMaterialFragment`?

pawelbaran (commented):
In general LGTM as for my limited understanding of the domain, but I left a few questions - please see the code comments.

Structure_oM/Results/Bar Results/BarRequiredArea.cs
IsakNaslundBh: Would agree with @pawelbaran here. m2/m != m for this case as it is different meters. Would add a new attribute, `[AreaPerUnitLength]`, something like:

```
    public class AreaPerUnitLength: QuantityAttribute
    {

        public override int L { get; } = 1;

        public override string SIUnit { get; } = "m²/m";

    }
```

Similar to https://github.com/BHoM/BHoM/blob/master/Quantities_oM/Attributes/Volume.cs vs https://github.com/BHoM/BHoM/blob/master/Quantities_oM/Attributes/SectionModulus.cs

Structure_oM/Results/Mesh/MeshRequiredArea.cs
peterjamesnugent: A reinforcement layout that follows the perimeter of the section, for example in a column you might want to specify it.

Structure_oM/Results/Bar Results/BarRequiredArea.cs
peterjamesnugent: To keep the result class light. For a single `Bar`, for 100 `Loadcase`s with 5 internal divisions you will have 500 result objects, then for 100s of `Bar`s you'd have the same `IMaterialFragment` stored over and over again. Similarly the `id` is stored rather than the `Bar` itself - we can then use post processing methods to get envelopes etc. and cross reference the `Material`, `Bar` etc.

Structure_oM/Results/Mesh/MeshRequiredArea.cs
pawelbaran: Yeah, that is why I got stuck on this one - I do not think I have ever seen the perimeter reinforcement in panel elements, but as above, I am nowhere close to any expertise in the field, leaning to you 100%.

Structure_oM/Results/Bar Results/BarRequiredArea.cs
pawelbaran: Hmm, but wouldn't it be possible to have only one `IMaterial` object in the memory and all `BarRequiredArea` objects storing just a reference to it? This should be safer as well as more structured - the question is whether this is possible. Possibly @IsakNaslundBh could help?

Structure_oM/Results/Mesh/MeshRequiredArea.cs
peterjamesnugent: On a particularly deep slab you might have reinforcement along the edge of the wall - but that is getting in to the `RebarIntent` territory (see sketch below). 

Given that this is representing the reinforcement at discrete points within a `Panel` I think it makes the most sense to remove it.

image

Structure_oM/Results/Mesh/MeshRequiredArea.cs
peterjamesnugent: Given that this is a high level representation I think it's appropriate as a simplification - otherwise we will need to start to define directions (which is covered in the `RebarIntent` representation). @IsakNaslundBh thoughts?

Structure_oM/Results/Bar Results/BarRequiredArea.cs
IsakNaslundBh: I was actually the one suggesting it to be set up the way @peterjamesnugent has done it, for the reasons he is mentioning above. Think it is important to keep the result classes as light as possible, and also would like them to easily be converted to table format for use in excel for example.

Storing one instance of the material in memory and have all results pointing to it is possible as long as the objects only stay in memory, in one process and only are accessed from this point, which is not the case for us. As soon as you would for example store them on mongo or similar, then that link is instantly gone, and you would end up with duplicates of that material again and again. Any for of serialisation of the objects, as well as anything requiring deep cloning of the object would have this impact.

Agree that care needs to be taken here, but still think this might be the best solution.

Structure_oM/Results/Mesh/MeshRequiredArea.cs
```diff
@@ -0,0 +1,87 @@
...
+    [Description("Minimum required area of reinforcement for an AreaElement.")]
+    public class MeshRequiredArea : MeshElementResult, IImmutable
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Area]
+        [Description("Minimum required area of reinforcement in the top layer.")]
+        public virtual double Top { get; }
```
IsakNaslundBh, line 41: Wonder if we would like to double up this? As in have TopX and TopY for the two main local directions?

Same for bottom reinforcement.

Also, think this should have the unit `AreaPerUnitLength` ...?

peterjamesnugent: @IsakNaslundBh that makes sense, and then the direction is dictated by the direction of the `Panel` that the result refers to, so it doesn't need to be stored in the `MeshRequiredArea` - but in terms of the class itself it does not need to deal with directions/regions or cuts.

Structure_oM/Results/Mesh/MeshRequiredArea.cs
peterjamesnugent: Just thinking about this a little more, I think having a `Shear` and `Torsion` required area at discrete nodes is useful, for example to demonstrate where additional reinforcement is required (e.g. high shears/torsion near boundaries that are not required elsewhere).

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check required

pawelbaran (approved):
LGTM now 👍

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check serialisation

pawelbaran:

IsakNaslundBh (approved):
Looks good now!

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check versioning
@BHoMBot check serialisation
@BHoMBot check core

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1255, opened 2021-07-06
Structure_oM: Add RebarIntent objects for Bars and Panels

### LINKED ISSUE #1241: Structure_oM: Rebar intent and Reinforcement Bar objects
#### Description:

Two levels here, I think the `RebarIntent` and `Rebar` objects should be linked by a common interface because they share similarities. 

1. Rebar Intent
- Inputs by users as a `List<Points>` and `List<ICurve>` and this allows flexibility in terms of multiple layers of longitudinal bars, and links etc. and maintains the existing `Longitudinal` and `Transverse` reinforcement (may require some changes to removing layouts)
- Allows the user to sketch the rebar intent and assign that to object
- If they want to use the `SectionProperty` from the `Bar` or `IAreaElement` we can write some Structure_Engine methods that basically offset perimeters etc. making use of some of the existing methods 
- This is owned by a `Bar` and `IAreaElement` as a `Fragment`

2. Rebar Detailing
- The `Rebar` makes use of the `IReinforcementBar` interface so inherits the properties that are common across the two representations
- Although it will have a spacing parameter and a shape code that it owns making it more refined 
- We need to be able to construct the various shape codes based on the `eNum` but right now the `ICurve` can define the bar whilst we map out various constructors
- The `RebarSet` is useful for downstream applications such as bundles of bars, but happy to use `tags` and other native BHoM properties to reduce additional object definitions 
- I think we should make some allowance for fire rating of the concrete, perhaps another fragment

image

@pawelbaran @al-fisher @IsakNaslundBh @alelom @rwemay

peterjamesnugent:

Closes #1241 

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/Versioning_Toolkit/%23168-ConcreteSectionUpgrade.gh?csf=1&web=1&e=YqA4Bo

### Changelog

- Removed `MinimumCover` attribute from `ConcreteSection` and migrated it to `RebarIntent`
- Changed how reinforcement is stored on a `ConcreteSection`, instead of `List<IBarReinforcement>` it now stores a single `BarRebarIntent` 
- Added a `BarRebarIntent` class that stores a `List<IBarReinforcement>` and the `MinimumCover` - this inherits the `IRebarIntent` interface as the overarching class for RebarIntent across different elements
- Added `IRebarIntent` interface
- Added `ReinforcementRegion` which defines the area of the `Panel` that is being reinforced, this is using an `ICurve` and a `Basis` 
- Added `PanelReinforcement` which stores the geometrical definition of the reinforcement, aligned to the `Orientation` of the `ReinforcementRegion` i.e. local x is longitudinal and local y is transverse
- `PanelReinforcement` also stores the `MinimumCover` to allow the different layers of reinforcement to have different covers
- Added `PanelRebarIntent` which inherits `IRebarIntent` and `IFragment` storing the reinforcement intent of the `Panel` and is added through the `Fragments`

### Additional comments

- Discussion around whether minimum cover should be stored on the `LongitudinalReinforcement` or `TransverseReinforcement` objects.
- Versioning to be added

As based on class diagram below:
Reinforcement


peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check compliance

Structure_oM/SectionProperties/Reinforcement/ReinforcementRegion.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace BH.oM.Structure.SectionProperties.Reinforcement
+{
+    [Description("A region defining the area of a Panel to be reinforced.")]
+    public class ReinforcementRegion : BHoMObject, IRegion
+    {
+        [Description("Local x, y, and z axes of the region as a vector Basis. Defaults to world axes.")]
+        public virtual Basis Orientation { get; set; } = Basis.XY;
+
+        [Description("A 2D curve defining the external boundaries of the region to be reinforced.")]
+        public virtual ICurve Perimeter { get; set; } = new Polyline();
```
IsakNaslundBh, line 37: Better stick to nulls as defaults rather than empty objects. Easier to check for.
```suggestion
        public virtual ICurve Perimeter { get; set; } = null;
```

Structure_oM/SectionProperties/Reinforcement/ReinforcementRegion.cs
```diff
@@ -0,0 +1,40 @@
...
+using BH.oM.Geometry;
+using System.ComponentModel;
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
+{
+    [Description("A region defining the area of a Panel to be reinforced.")]
+    public class ReinforcementRegion : BHoMObject, IRegion
+    {
+        [Description("Local x, y, and z axes of the region as a vector Basis. Defaults to world axes.")]
+        public virtual Basis Orientation { get; set; } = Basis.XY;
```
IsakNaslundBh, line 34: Would change this to null as well, but for a different reason. For this, I think we are better of using the panel coordinate system as default if nothing is provided here, i.e. if this is null.

```suggestion
        public virtual Basis Orientation { get; set; } = null
```

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
```diff
@@ -0,0 +1,66 @@
...
+        [Description("The region defining the area of the Panel to be reinforced.")]
+        public virtual ReinforcementRegion Region { get; set; }
+
+        [Description("The diameter of the longitudinal reinforcement.")]
+        public virtual double LongitudinalDiameter { get; set; }
+
+        [Description("The spacing of longitudinal reinforcement.")]
+        public virtual double LongitudinalSpacing { get; set; }
+
+        [Description("The depth of longitudinal reinforcement measured from the centre of Panel. This will be negative for bottom reinforcement.")]
```
IsakNaslundBh, line 49:
```suggestion
        [Description("The depth of longitudinal reinforcement measured from the centre of Panel, along the Normal. This will be negative for bottom reinforcement.")]
```

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
```diff
@@ -0,0 +1,66 @@
...
+using BH.oM.Base;
+using BH.oM.Structure.MaterialFragments;
+using BH.oM.Spatial.Layouts;
+using BH.oM.Quantities.Attributes;
+using System.ComponentModel;
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
+{
+    [Description("Defines the  reinforcement of a Panel in the longitudinal and transverse direction specified by the Basis of the ReinforcementRegion.")]
+    public class PanelReinforcement : BHoMObject
```
IsakNaslundBh, line 32: All numeric properties need quantity attributes (Length).

Also, wonder if we want something controlling if reinforcement in a specific direction should exist or not. Could either be via some convention, such as diameter in that direction == 0 or similar, or via some controlling booleans.

If we use some convention, I think the description needs to highlight it.

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
```diff
@@ -0,0 +1,44 @@
...
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
+{
+    [Description("A collection of reinforcement for a Panel.")]
+    public class PanelRebarIntent : BHoMObject, IFragment, IRebarIntent
```
IsakNaslundBh, line 32: A question more than anything. On the Bars we have the minimum cover on the `BarRebarIntent`, while here it sits on the `PanelReinforcement`. WOuld it be possible to move the MinimumCover to the `PanelRebarIntent` instead?

IsakNaslundBh (dismissed):
Some initial comments

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
peterjamesnugent: I pondered this in the additional comments.

For a Panel it makes sense that each `PanelReinforcement` has it's own `MinimumCover`, because the top reinforcement may have a different exposure that the bottom reinforcement - or the inside/outside of a wall. 

So I wondered whether for the `BarRebarIntent` we should actually put it on the `LongitudinalReinforcement` and `TransverseReinforcement` as they are effectively analogous with the `PanelReinforcement`.

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
peterjamesnugent: I think the diameter being set to 0 would be simplest - because any follow on methods that visualise the bar or do any sort of calculation will use the diameter. I'll update the desc.

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
```diff
@@ -0,0 +1,66 @@
...
+        [Description("The diameter of the longitudinal reinforcement.")]
+        public virtual double LongitudinalDiameter { get; set; }
+
+        [Description("The spacing of longitudinal reinforcement.")]
+        public virtual double LongitudinalSpacing { get; set; }
+
+        [Description("The depth of longitudinal reinforcement measured from the centre of Panel. This will be negative for bottom reinforcement.")]
+        public virtual double LongitudinalDepth { get; set; }
+
+        [Description("The diameter of the transverse reinforcement.")]
```
IsakNaslundBh, line 52: To add as well, this description does not really add any value. What you have in the Create method in the Engine is far superior. Would simply switch all descriptions here to what you have in the engine.

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
IsakNaslundBh: Cool. Would put it on both of the Diameter property descriptions as well

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
IsakNaslundBh: That is also an option..

Another option would be to have two minimum cover properties on here, `MinmumCoverTop` and `MinimumCoverBottom`. Then you could also describe what top means (along positive direction of the normal) vs Bottom (the opposite). But, not sure if that is better.

peterjamesnugent: I would be hesitant to adopt top/bottom because what if it's a wall, does it become left/right?

The way I see it working is that if a `Panel` were to have top and bottom reinforcement, the `PanelRebarIntent` object would own a `PanelReinforcement` object summarising the top layer diameter etc. and a second `PanelReinforcement` object summarising the bottom layer, each with their own `MinimumCover` so we don't have to worry about orientations or conventions for top/bottom/left/right.

peterjamesnugent: Just to add to this, after discussion with @IsakNaslundBh, whilst a `Bar` can have different covers on it's different faces the reinforcement itself will typically reinforce many faces. Therefore, capturing this information at the RebarIntent level feels like a level of detail too far given that it is a design intent. This is an aspect that can be captured in the full detail of the reinforcement in their 3D representation.

To summarise:
- `BarRebarIntent` to store the `MinimumCover` applicable to the whole `Bar`
- `PanelReinforcement` to store a `Minimum` cover so one `Panel` can have many `MinimumCover` properties

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
```diff
@@ -0,0 +1,44 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
```
pawelbaran, line 29: This namespace is suspicious? Same refers to the file location.

Structure_oM/SectionProperties/Reinforcement/BarRebarIntent.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    [Description("A collection of reinforcement for a Bar and minimum cover.")]
+    public class BarRebarIntent : BHoMObject, IRebarIntent
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("A collection of IBarReinforcement objects.")]
+        public virtual List<IBarReinforcement> BarReinforcement { get; set; }
```
pawelbaran, line 39: Usually the properties with a setter are initialised in the declaration. This is particularly helpful in case of collections, to avoid unexpected nulls. Not sure if this is just an observation or a solid recommendation, maybe @FraserGreenroyd could tell more?
```suggestion
        public virtual List<IBarReinforcement> BarReinforcement { get; set; } = new List<IBarReinforcement>();
```

Structure_oM/SectionProperties/Reinforcement/IRebarIntent.cs
```diff
@@ -0,0 +1,35 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+
+using System.ComponentModel;
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
```
pawelbaran, line 27: This namespace (as well as file location) is suspicious

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    [Description("A collection of reinforcement for a Panel.")]
```
pawelbaran, line 31: I would consider elaborating a bit more here, to highlight that this is intent level, not the full reinforcement information

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class PanelRebarIntent : BHoMObject, IFragment, IRebarIntent
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("A collection of PanelReinforcement objects")]
```
pawelbaran, line 38: I know this may sound mean, but shouldn't all descriptions end with a full stop? I do not mind, but @FraserGreenroyd @IsakNaslundBh may have a strong opinion on that 😃

Structure_oM/SectionProperties/Reinforcement/PanelReinforcement.cs
```diff
@@ -0,0 +1,76 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Structure.MaterialFragments;
+using BH.oM.Spatial.Layouts;
+using BH.oM.Quantities.Attributes;
+using System.ComponentModel;
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
```
pawelbaran, line 29: As above, I would consider moving the general concept of reinforcement to some more suitable namespace and folder.

Structure_oM/SectionProperties/Reinforcement/ReinforcementRegion.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace BH.oM.Structure.SectionProperties.Reinforcement
+{
+    [Description("A region defining the area of a Panel to be reinforced.")]
+    public class ReinforcementRegion : BHoMObject, IRegion
+    {
+        [Description("Local x, y, and z axes of the region as a vector Basis. Defaults to world axes.")]
+        public virtual Basis Orientation { get; set; } = null;
+
+        [Description("A 2D curve defining the external boundaries of the region to be reinforced.")]
+        public virtual ICurve Perimeter { get; set; } = null;
```
pawelbaran, line 37: What would happen if the perimeter does not fully overlap with the outline of the panel? Would the Boolean intersection be taken? Going further this way, what if the perimeter lies in different plane than the panel? Should it be projected then or error thrown? May be worth explaining this a bit in the description, I would also consider the consequences of different approaches to make sure we are not turning towards a dead end.

pawelbaran (changes requested):
I have left a few comments and questions - please see the code comments.

Besides that, if you zoom out, it feels a bit weird to have 2 different approaches to how we structure our object model: that the panel rebar intent sits inside a fragment (composition), while in case of bars it has a dedicated property in concrete subtype (quasi-inheritance). I am aware it has been discussed widely and I was one of the most active participants, but I still feel a bit uncomfortable with such misalignment. Having said that, I am not sure if I can propose anything better 😉

pawelbaran:
Actually, following on my comment on the Engine, I think this PR is missing versioning of properties that are changed or deleted.

FraserGreenroyd:
@BHoMBot check versioning

Structure_oM/SectionProperties/Reinforcement/PanelRebarIntent.cs
peterjamesnugent: Changed to `BH.oM.Structure.SurfaceProperties.Reinforcement`.

I've left the IRebarIntent in the `BH.oM.Structure.SectionProperties.Reinforcement` because I can't think of a better place to put it.

Structure_oM/SectionProperties/Reinforcement/IRebarIntent.cs
peterjamesnugent: Moved to `BH.oM.Structure`, so it sits alongside `IProperty` which is also another base interface which is inherited by a number of objects across different sub namespaces within `BH.oM.Structure`.

Structure_oM/SectionProperties/Reinforcement/ReinforcementRegion.cs
peterjamesnugent: I've modified the description - I think it should always exist within the `Panel` and if it doesn't we raise an error (this can explained/implemented on the Engine side).

FraserGreenroyd:
@BHoMBot check versioning

Structure_oM/SurfaceProperties/Reinforcement/PanelRebarIntent.cs
```diff
@@ -0,0 +1,45 @@
...
+    [Description("A collection of PanelReinforcement objects that represent the rebar intent. \n" +
+        "This is stored as layers of reinforcement amongst regions within the Panel.")]
+    public class PanelRebarIntent : BHoMObject, IFragment, IRebarIntent
+    {
+        /***************************************************/
+        /**** Properties                                ****/
+        /***************************************************/
+
+        [Description("A collection of PanelReinforcement objects.")]
+        public virtual List<PanelReinforcement> PanelReinforcement { get; set; }
```
pawelbaran, line 40: I would initialise the list here to avoid unexpected nulls.
```suggestion
        public virtual List<PanelReinforcement> PanelReinforcement { get; set; } = new List<PanelReinforcement>();
```

Structure_oM/SurfaceProperties/Reinforcement/ReinforcementRegion.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace BH.oM.Structure.SurfaceProperties.Reinforcement
+{
+    [Description("A region defining the area of a Panel to be reinforced. The region must exist within the ExternalEdges of the Panel.")]
+    public class ReinforcementRegion : BHoMObject, IRegion
+    {
+        [Description("Local x, y, and z axes of the region as a vector Basis. Defaults to world axes.")]
+        public virtual Basis Orientation { get; set; } = null;
+
+        [Description("A 2D curve defining the external boundaries of the region to be reinforced. The curve must exist within the ExternalEdges of the Panel.")]
```
pawelbaran, line 36:
```suggestion
        [Description("A 2D curve defining the external boundaries of the region to be reinforced. The curve must exist within the 2D region outlined by ExternalEdges of the Panel.")]
```

pawelbaran (changes requested):
Looks good in general, although I found one minor issue plus made one suggestion on the descriptions - please see the code comments.

Besides that, I am wondering about the namespace naming: atm `PanelRebarIntent` and `PanelReinforcement` sit in `BH.oM.Structure.SurfaceProperties.Reinforcement`, which might be a bit misleading, taken that the reinforcement is not related to the properties of a panel. As an alternative, maybe we should have an entirely separate namespace for rebar, e.g. `BH.oM.Structure.Reinforcement`? With subnamespaces for 1D and 2D or similar. This could help us avoid similar considerations in the future. What do you think, @peterjamesnugent? Just in case, this can be easily updated in the future using Versioning, so I am not going to block the PR because of that.

Structure_oM/IRebarIntent.cs
```diff
@@ -0,0 +1,35 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+
+using System.ComponentModel;
+
+namespace BH.oM.Structure
```
pawelbaran, line 27:
```suggestion
namespace BH.oM.Structure.Reinforcement
```

Structure_oM/SectionProperties/Reinforcement/BarRebarIntent.cs
```diff
@@ -0,0 +1,48 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+
+namespace BH.oM.Structure.SectionProperties.Reinforcement
```
pawelbaran, line 29:
```suggestion
namespace BH.oM.Structure.Reinforcement
```

Structure_oM/SectionProperties/Reinforcement/IBarReinforcement.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace BH.oM.Structure.SectionProperties.Reinforcement
```
pawelbaran, line 28:
```suggestion
namespace BH.oM.Structure.Reinforcement
```

pawelbaran (changes requested):
I would put all objects related to reinforcement in `BH.oM.Structure.Reinforcement` namespace. 

Possibly the above could be further split into subnamespaces to account for dimensionality (bar vs panel) - this may however become counterproductive in case of objects applicable to both, so I would rather wait with it until we are confident.

What do you think @peterjamesnugent @IsakNaslundBh?

Structure_oM/IRebarIntent.cs
peterjamesnugent: @pawelbaran my only apprehension to this is that it doesn't reflect the structure of the properties. The `BarRebarIntent` is stored on the ConcreteSection which is why it currently sits under `BH.oM.Structure.SectionProperties.Reinforcement`. Although I do see the simplicity of everything being under a a single name space.

I don't think we should have a split between 1D/2D or Bar/Panel under the `Reinforcement` namespace because the objects are generally bespoke (i.e. related to the object) - so it would end up with unnecessarily deep namespaces without much benefit.

pawelbaran: > @pawelbaran my only apprehension to this is that it doesn't reflect the structure of the properties. The `BarRebarIntent` is stored on the ConcreteSection which is why it currently sits under `BH.oM.Structure.SectionProperties.Reinforcement`. Although I do see the simplicity of everything being under a a single name space.

Hmm makes sense, I could agree with that. However, as a reinforcement newbie I think I would intuitively expect everything related to that topic sitting in one folder and namespace (now it is spread across 3). This is just my opinion though, and if you and @IsakNaslundBh are happy with the current option, I am happy to approve 😉

IsakNaslundBh: Would agree with mvong it all into `BH.oM.Structure.Reinforcement` namespace/folder. Would probably not put it into any deeper namespace though as mentioned by @peterjamesnugent above. Unless it gets completely out of hand in terms of number of classes or impossible to set unambiguous names, and think neither of those will be of any issue.

peterjamesnugent: Just discussed with @IsakNaslundBh - there's quite a few instances where properties are not reflected in namespace organisation (such as `Material` being owned by a `SectionProperty` and not sitting under that namespace - so I was probably overthinking the namespace for the `Reinforcement`. 

I'll put everything under a single `Reinforcement` namespace.

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check versioning

Structure_oM/IRebarIntent.cs
pawelbaran: Sorry for being a PITA, but I would consider putting `IRebarIntent` in `BH.oM.Structure.Reinforcement` namespace too, for the sake of consistency (same with folder structure ofc). Besides that I am 100% happy with the PR 👍

pawelbaran (approved):
LGTM now 👍

pawelbaran:
@BHoMBot check required

IsakNaslundBh (approved):
Looks good to be now!

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1294, opened 2021-10-01
Base_oM: Adding Enumeration object

### LINKED ISSUE #1293: Adding Enumeration object that can implement an interface
#### Description:

Something like this is not possible with simple enums:

image

See [this link](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types) for an example of implementation

adecler:

Closes #1293

See issue. Current implementation in the context of the Embodied Carbon toolkit

### Test files
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM/Base_oM/%231294%20-%20Enumeration?csf=1&web=1&e=eX3XoG
You will need the following PRs to do the testing:
- https://github.com/BHoM/BHoM_Engine/pull/2637
- https://github.com/BHoM/Excel_Toolkit/pull/310
- https://github.com/BuroHappoldEngineering/EmbodiedCarbon_Toolkit/pull/1

### Additional comments
There's a lot that can be done to make the Enumerations more user friendly in the UI (e.g. make it behave exactly like a regular enum in the UI) but this PR is just about laying the foundation for this new type of object. It has been working great for me in the context of the Carbon calculator tool for all stages of workflow so I am pretty confident it is safe to merge. Make sure to play with the test file though.

### DIFF
--- a/BHoM/Enumeration.cs
+++ b/BHoM/Enumeration.cs
@@ -27,7 +27,7 @@ using System.Reflection;
 
 namespace BH.oM.Base
 {
-    public abstract class Enumeration : IObject, IComparable, IEnum // TODO: Fix serialisation so we can remove the inheritance from BHoMObject
+    public abstract class Enumeration : IObject, IEnum 
     {
         /***************************************************/
         /**** Properties                                ****/
@@ -89,9 +89,9 @@ namespace BH.oM.Base
 
         /***************************************************/
 
-        public int CompareTo(object other)
+        public int CompareTo(IEnum other)
         {
-            return Value.CompareTo(((Enumeration)other)?.Value);
+            return Value.CompareTo(other?.Value);
         }
 
 
--- a/BHoM/Enumeration.cs
+++ b/BHoM/Enumeration.cs
@@ -91,7 +91,7 @@ namespace BH.oM.Base
 
         public int CompareTo(object other)
         {
-            return Value.CompareTo(((Enumeration)other).Value);
+            return Value.CompareTo(((Enumeration)other)?.Value);
         }
 
 
--- a/BHoM/Enumeration.cs
+++ b/BHoM/Enumeration.cs
@@ -1,0 +1,132 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2021, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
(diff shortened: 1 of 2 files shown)


BHoM/Enumeration.cs
```diff
@@ -0,0 +1,132 @@
...
+        /***************************************************/
+
+        public static IEnumerable<T> GetAll<T>() where T : Enumeration
+        {
+            return typeof(T).GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly)
+                     .Select(f => f.GetValue(null))
+                     .Cast<T>();
+        } 
+
+        /***************************************************/
```
IsakNaslundBh, line 127: As discussed, these methods really should live in the Engine. Ok for that to happen in a subsequent PR, but if an issue can be raised to capture that!

BHoM/Interface/IEnum.cs
```diff
@@ -0,0 +1,54 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace BH.oM.Base
+{
+    public interface IEnum : IComparable
```
IsakNaslundBh, line 29: Question here, not really a change request, but any reason why this is not implementing `IComparable<IEnum>` instead of for object? Also, with the additional override methods required, could it not also implement `IEquatable<IEnum>` ? Seems like a close overlap.

IsakNaslundBh (commented):
Things work as expected with the provided test file.

Have left some comments below with some questions/comments.

BHoM/Interface/IEnum.cs
adecler: I have switched to `IComparable<IEnum>`.
I haven't added `IEquatable<IEnum>` because it requires an additional `Equals(IEnum)` and I don't see the point of that. It's not like we are using it anywhere else either.

BHoM/Enumeration.cs
adecler: I am not too sure about that.

`Enumeration.GetAll<T>()` would translate into something like `BH.oM.Base.Query.Values<T>()` and therefore give a component like  

image

One key point of separating the methods into the Engine is to provide a matching functionality in the UI but I'd rather not have that `Values` component in the UI as it makes absolutely no sense.

On a similar vein (although I am a bit more hesitant on this one), I prefer `static bool Equals(IEnum a, IEnum b)` to be in the object definition. The whole purpose of this method compared to the non-static alterative is that it allows to make it more readable in the code that we are comparing Enumerations by having a clear call to `Enumeration.Equals`. If we take that away by moving it to an Engine method `Query.Equals`, we might as well remove that method altogether. Admittedly, there isn't much reason right now to keep it so happy to do just that. It could become handy if we add implicit casting from string (something I would like to add later) though.

IsakNaslundBh (approved):
Approving after discussion abut the below points.

adecler:
@BHoMBot check required

FraserGreenroyd (commented):
Have discussed with @adecler to raise an issue for `virtual` keyword compliance and merge this PR as is with dispensation on that check, with the `virtual` stuff being fixed in a smaller PR.

adecler:
@FraserGreenroyd , I have raised teh issue about the missing virtual keyword here: https://github.com/BHoM/BHoM/issues/1324
Will get on it tomorrow

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 4222680812

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #1298, opened 2021-10-06
BHoM: ISettings to implement IObject

### LINKED ISSUE #1297: BHoM: ISettings to be an IObject
#### Broken rules:

Currently the `ISettings` interface does not inherit from `IObject`. This can lead to some issues in reflection and serialiser if one is trying to use an object that only implements the interface, but no other interface that leads to IObject.

Is there any reason why ISettings is not an IObejct?

IsakNaslundBh:

Closes #1297

Enables things like serialisation for objects that _only_ implements the `ISettings` interface, without also being a `BHoMObject`.

Should not change the behaviour of the pre-existing `ISettings` as they all currently are `BHoMObjects` and get the `IObject` from there.

### DIFF
--- a/BHoM/Interface/ISettings.cs
+++ b/BHoM/Interface/ISettings.cs
@@ -31,7 +31,7 @@ using System.Threading.Tasks;
 namespace BH.oM.Base
 {
     [Description("Toolkit Settings that need to be saved permanently.")]
-    public interface ISettings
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

PR #1299, opened 2021-10-06
BHoM: Add and refactor objects for Diffing

### LINKED ISSUE #1295: BHoM: add ComparisonFunctions property and class for the ComparisonConfig object
#### Description:

In order to support more flexibility for the Diffing and Hash algorithms, we can provide delegates that are to be invoked during their computation.

These delegates can be any function defined in downstream Toolkits. This allows to support particular comparison scenarios that may arise, e.g. in Revit_Toolkit we want to treat "RevitParameter" names as they were object declared properties.

alelom:

Closes #1295


alelom:
@BHoMBot check required

PR #1360, opened 2022-02-07
Analytical_oM: IResultCollection generic update

### LINKED ISSUE #1359: Analytical_oM: Consider removal of non-generic IResultCollection

We currently have a generic `IResultCollection` as well as a `IResultCollection<T>`. Should investigate the usefullness of keeping the non-generic interface, or if it can be removed.

IsakNaslundBh:

Closes #1359 

Changing property type from `ReadOnlyCollection` to `IReadOnlyList` which allows for the use of the covariant `out` keyword on the interface. This means that an engine method written to accept for example an `IResultCollection<IResult>` should accept an `IResultCollection<BarForce>` similar to the way `IEnumerable<object>` etc works.

Also will give similar benefits to the `IMeshResult<T>` interface which also will be able to use the out keyword.

With this change the non-generic class can safely be removed as the only reason for this interface was to help with filtering.

### Additional comments

Some toolkits will need slight alignment before this is merged. Raising PR for @alelom to have a first look.

### DIFF
--- a/Analytical_oM/Results/IMeshResult.cs
+++ b/Analytical_oM/Results/IMeshResult.cs
@@ -30,7 +30,7 @@ using System.ComponentModel;
 namespace BH.oM.Analytical.Results
 {
     [Description("Base interface for any Mesh result class which is a collection of discrete MeshElementResults.")]
-    public interface IMeshResult<T> : IObjectIdResult, IResultCollection<T> where T : IMeshElementResult
+    public interface IMeshResult<out T> : IObjectIdResult, IResultCollection<T> where T : IMeshElementResult
     {
 
     }
--- a/Analytical_oM/Results/IResultCollection.cs
+++ b/Analytical_oM/Results/IResultCollection.cs
@@ -27,17 +27,6 @@ using System.ComponentModel;
 
 namespace BH.oM.Analytical.Results
 {
-
-    /***************************************************/
-
-    [Description("Non-generic base interface for result collections. The IResultCollection facilitates simpler filtering in the Engine. Classes are generally recommended to implement the IResultCollection<T> interface.")]
-    public interface IResultCollection : IObject, IImmutable, IResult
-    {
-
-    }
-
-    /***************************************************/
-
     [Description("Base interface for results classes that are a collection of individual results, for example IMeshResults.")]
     public interface IResultCollection<out T> : IResult, IImmutable where T : IResult
     {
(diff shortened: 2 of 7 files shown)


IsakNaslundBh:
@BHoMBot check installer

adecler:
@IsakNaslundBh , have you made sure the serialisation and versioning work for this new type of generic ?

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
> @IsakNaslundBh , have you made sure the serialisation and versioning work for this new type of generic ?

Actually had not double checked that with the serialisation check, but I have opened scripts with internalised MeshResults that have worked just fine, so assumed from that serialisation was ok, but looks like that is not the case for some reason according to the check. Will investigate.

IsakNaslundBh:
@BHoMBot check serialisation

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check compliance

alelom:
I think that https://github.com/BHoM/BHoM/pull/1360/commits/267b030263bda9b5eac1cc9309ba3d14baf4906b is an acceptable workaround to allow serialisation to work for the time being. Probably worth investigating why this is needed.

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
> I think that [267b030](https://github.com/BHoM/BHoM/commit/267b030263bda9b5eac1cc9309ba3d14baf4906b) is an acceptable workaround to allow serialisation to work for the time being. Probably worth investigating why this is needed.

Think it is required as the serialisation engine does not store the type of the collection being serialised out. It treats any list-like collection the same, by essentially just creating a JSON array, and then relies on the property type to decide what to de-serialise the object back to. With the change of the property type to an interface, I think (have not been able to exactly verify this) it simply creates a list rather than a a `ReadOnlyCollection<T>` which then did not match what was reuquired by the constructor.

Agree to some extent that this could be seen as a workaround, but also think changing the requirement on the constructor to `IEnumerable<T>`  makes it easier to call the constructor from the outside, and lets the class itself deal with making itself immutable. One could argue that this puts logic in the oM rather than the engine, but on this level I personally at least think the slightly elongated line with a null check, constructor call and ToList call is small enough for this to be fine.

IsakNaslundBh:
@BHoMBot check required

alelom (approved):
I like this solution! Tested with the Engine PR script, works.

Link to one of my previous comments for reference: https://github.com/BHoM/BHoM/pull/1360#issuecomment-1050064853

LGTM!

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1383, opened 2022-04-21
Analytical o m #1380 add identifier categorisation attribute for results

### LINKED ISSUE #1380: Analytical_oM: Add identifier categorisation attribute for results
#### Description:

Add ways to categorise what a identifier targets, to distinguish between Scenario Identifiers and Object identifiers. This will allow for easier and more generic ways of grouping result depending on which identifers they contain

see 
image

for example

IsakNaslundBh:

Closes #1380

### DIFF
--- a/Analytical_oM/Results/Attributes/IdentifierAttribute.cs
+++ b/Analytical_oM/Results/Attributes/IdentifierAttribute.cs
@@ -1,0 +1,35 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.ComponentModel;
+
+namespace BH.oM.Analytical.Results
+{
+    [Description("Base attribute for identifier attributes for properties of results.")]
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
+    public abstract class IdentifierAttribute : Attribute
+    {
+    }
+}
--- a/Analytical_oM/Results/Attributes/ObjectIdentifierAttribute.cs
+++ b/Analytical_oM/Results/Attributes/ObjectIdentifierAttribute.cs
@@ -1,0 +1,36 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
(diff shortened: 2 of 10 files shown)


IsakNaslundBh:
@BHoMBot check compliance

alelom (dismissed):
Tested both provided scripts. The scripts were clear and behaved as expected.

The code also looks good and the reasons behind this PR are clear. 

Only few minor comments:
- I suggested some changes to the descriptions of the attributes to make them clearer from my perspective.
- I am not sure I like the word "scenario", I might prefer _context_ instead but I am not too sure either. Would be good to have others' opinion, but I do not think this should be a blocker to this PR.
- Just as a note, this PR expands the current BHoM definition of the term "Identifiers". This is in particular with reference to the Wiki and how we mention AdapterId/BHoM_Guid. I initially thought that I would prefer avoiding using the term "Identifiers" for the properties/group of properties highlighted in this PR, and default to a more generic _properties_. However, @IsakNaslundBh convinced me otherwise as these are in fact properties that identify an object and possibly link it to other objects. Just as a note, we might have to be clear on this in future rewritings of the wiki, possibly adding a new page containing the diagram present in https://github.com/BHoM/BHoM/issues/1380.

IsakNaslundBh:
Thanks @alelom !

> * I suggested some changes to the descriptions of the attributes to make them clearer from my perspective.
Fully agree with your changes. Have pushed your suggested changes 👍 

> * I am not sure I like the word "scenario", I might prefer _context_ instead but I am not too sure either. Would be good to have others' opinion, but I do not think this should be a blocker to this PR.
I personally like the Scenario word, but I am not fully attached to it and would not scream and shout if it was changed either 😄 .
I do not dislike "Context" but I kind of feel Scenario is more precise (but might be convinced otherwise!). The more I think about it, the less certain I am between the two, with a slight edge to scenario still. Now, I am just rambling, but, Happy with either really.

> * Just as a note, this PR expands the current BHoM definition of the term "Identifiers". This is in particular with reference to the Wiki and how we mention AdapterId/BHoM_Guid. I initially thought that I would prefer avoiding using the term "Identifiers" for the properties/group of properties highlighted in this PR, and default to a more generic _properties_. However, @IsakNaslundBh convinced me otherwise as these are in fact properties that identify an object and possibly link it to other objects. Just as a note, we might have to be clear on this in future rewritings of the wiki, possibly adding a new page containing the diagram present in #1380.

Yep, agree. As stated offline for context, I personally kind of like the overlap in nomenclature, as it is fundamentally taking about the same thing. ObjectId on say a BarForce will always correspond to a AdapterId/BHoM_Guid on a Bar for example, hence I think there is even some power in keeping it as is! :)

If you do not mind re-approving now that I have pushed the changes, and also approve the Engine (if you are happy with it), I might get this merged! :)

FraserGreenroyd:
> > * I am not sure I like the word "scenario", I might prefer _context_ instead but I am not too sure either. Would be good to
> >   have others' opinion, but I do not think this should be a blocker to this PR.
> 
> I personally like the Scenario word, but I am not fully attached to it and would not scream and shout if it was changed either 😄 . I do not dislike "Context" but I kind of feel Scenario is more precise (but might be convinced otherwise!). The more I think about it, the less certain I am between the two, with a slight edge to scenario still. Now, I am just rambling, but, Happy with either really.


For those coming into this new to provide such opinions - can we get some context as to the scenario which has precipitated this discussion?

IsakNaslundBh:
> For those coming into this new to provide such opinions - can we get some context as to the scenario which has precipitated this discussion?

Yes, sorry. Some context on the issue, but the general intention is to be able to identify which properties on the result are intended to help identify the scenario/context of the result.

Could for example for a structural result be a result with a particular `timestep` from a particular `Loadcase` that had been analysed, or a Particular mode from a particular Loadcase, or for most common cases, just the ResultCase by itself.

For something else this could for example be all identifiers required to state that the result corresponds to mid day, sunny day in June or something similar.

Basically all identification properties required to be able to specify what the circumstances where when the result was generated, while the counterpart ObjectIdentifiers are instead used to identify the object, or subpart of the object, the result belong to.

Not sure if this makes it clearer?

alelom (approved):

alelom:
As stated in my previous review, I am approving on the basis of the Scripts, the changes made by @IsakNaslundBh and his answers to my comments. Further discussions on terminology may be continued independently from this PR.

FraserGreenroyd:
> For something else this could for example be all identifiers required to state that the result corresponds to mid day, sunny day in June or something similar.
> 
> Basically all identification properties required to be able to specify what the circumstances where when the result was generated, while the counterpart ObjectIdentifiers are instead used to identify the object, or subpart of the object, the result belong to.
> 
> Not sure if this makes it clearer?

This does make it clearer (clearer than my poor attempt at humour does at any rate 😉 ).

My opinion then, which I will caveat by saying it is just an opinion based on my use of the English language (which is by no means perfect nor likely to be correct by any stretch of the imagination!), would be that `Context` would be a more appropriate word to describe what you've described in the bit I've started the quote with.

My reasoning for this, is that to me, the scenario is we might be simulating load cases across the whole of summer. The context for this specific load case is that it was a sunny day in June. This context is different from a rainy day in July, but both contexts are part of the same scenario.

So for me, the scenario is the bigger/broader picture encapsulation of the task, with context being the more specific details on an individual result.

Obviously I have not read through this body of work in full, so it could be that `Scenario` is the right word based on what it is doing, so I could have completely the wrong end of the stick! But that would be my tuppence worth of opinion on the topic for now 😄

alelom:
@FraserGreenroyd I agree with your point of view which is the same where I'm coming from.

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check core
@BHoMBot check serialisation
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 6156887135

Dispensing due to known issues with check ready to merge failing to pick up successful run of check core

PR #1389, opened 2022-05-13
Graphics_oM: Add graphics render texture

### LINKED ISSUE #485: Graphics_oM : Add a Graphics.MaterialFragments.IMaterialFragment
This can follow the work restructuring the `Physical.Materials.Material` and `MaterialFragments` in `BH.oM.Structure` and `BH.oM.Environment` 

This `Graphics.MaterialFragment` can contain material render properties, translucency models etc.

IsakNaslundBh:

Closes #485

Add graphics texture class to handle rendering of surface and meshes. Defaults to BHoM Pink with no bitmap.

Adding a explicit casting from Colour to Texture to allow easier generation from the UI.

General intended logic of the class is that the Colour property always gives the colour of the edges, while the Texture property gives the colour or texture of the mesh/surface faces. If the texture is not set, the display should fall back on using the colour property instead.

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM/Graphics_oM/%231389-AddGraphsicsRenderTexture?csf=1&web=1&e=NxVT3M

### Additional comments

- Can also add transparency and bumb maps if we like, but kept it simple, only adding main texture in the first push.
- Later PR to handle adding a IMaterialProperty with a texture that can be applied to a physical Material class.

### DIFF
--- a/Graphics_oM/Render/RenderGeometry.cs
+++ b/Graphics_oM/Render/RenderGeometry.cs
@@ -42,8 +42,8 @@ namespace BH.oM.Graphics
         [Description("Colour used to render the edges of the Geometry. If texture is not set, this colour will also be used to render any meshes and surfaces of the Geometry. Default is BHoM Coral with a subtle transparency (Color.FromArgb(80, 255, 41, 105)).")]
         public virtual Color Colour { get; set; } = Color.FromArgb(80, 255, 41, 105);
 
-        [Description("Texture to be used for mesh and surface geometry. If null, the colour will be used for mesh and surface rendering instead.")]
-        public virtual Texture Texture { get; set; } = null;
+        [Description("Optional texture or colour to be used for mesh and surface geometry. If null, the colour will be used for mesh and surface rendering instead.")]
+        public virtual Texture SurfaceColour { get; set; } = null;
 
         [Description("The thickness of the curve. Should be a value greater or equal to 1.")]
         public virtual int EdgeThickness { get; set; } = 1;
--- a/Graphics_oM/Render/RenderGeometry.cs
+++ b/Graphics_oM/Render/RenderGeometry.cs
@@ -39,8 +39,14 @@ namespace BH.oM.Graphics
         [Description("A geometry (or many geometry objects collected into a single `CompositeGeometry` object).")]
         public virtual IGeometry Geometry { get; set; }
 
-        [Description("Colour used to render the Geometry. Default is BHoM Coral with a subtle transparency (Color.FromArgb(80, 255, 41, 105)).")]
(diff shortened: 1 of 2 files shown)


Graphics_oM/Render/RenderGeometry.cs
```diff
@@ -39,9 +39,15 @@ public class RenderGeometry : IRender
         [Description("A geometry (or many geometry objects collected into a single `CompositeGeometry` object).")]
         public virtual IGeometry Geometry { get; set; }
 
-        [Description("Colour used to render the Geometry. Default is BHoM Coral with a subtle transparency (Color.FromArgb(80, 255, 41, 105)).")]
+        [Description("Colour used to render the edges of the Geometry. If texture is not set, this colour will also be used to render any meshes and surfaces of the Geometry. Default is BHoM Coral with a subtle transparency (Color.FromArgb(80, 255, 41, 105)).")]
         public virtual Color Colour { get; set; } = Color.FromArgb(80, 255, 41, 105);
 
+        [Description("Optional texture or colour to be used for mesh and surface geometry. If null, the colour will be used for mesh and surface rendering instead.")]
+        public virtual Texture SurfaceColour { get; set; } = null;
```
michaelhoehn, line 46: Could this just be called Texture? Seems like specifying geometry type in the name could be too specific.

IsakNaslundBh: Actually think I just called it texture at first, but after a few back and forwards landed in this name for a few reasons. (rightly or wrongly, happy for your input if you think it makes sense or not)

- Firstly, what we wanted to achieve on this class is the ability to have 2 different colours, one for mesh/surface geometry and one for edges.
- The `RenderGeometry` is implementing the `IRender` Interface, and the `IRender` interface has a `Colour` property enforced. This means we need this class to have a property called `Colour`.
- For the Surface/Mesh rendering we want to enable more functionality and control (through the texture object)
- I have implemented an explicit convert from Colour -> Texture that makes in possible to plug in a colour to the texture object from the UIs.

With all this, we basically are enforced to have a Colour property, of type `System.Drawing.Color` and as we want to enable more options for the surface/mesh redering, it makes most sense to be the one to use for the Edges.
As the Property simply called "Colour" is used for the edges, we want to distinguish the property used for surfaces/meshes, and also highlight that it is possible to simply plug in a colour from the UI, which led to the name SurfaceColour.

If this does not make any sense, happy to revise but felt was worth trying to explain how the name ended up to what it did.

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 6622617384

PR #1423, opened 2022-08-25
Physical_oM: Add MaterialTakeoff class

### LINKED ISSUE #1422: Physical_oM: Change to methods required for IElementM
Given support for this given in discussion as well as offline calls, turning this into issue to be implemented!

### Discussed in https://github.com/orgs/BHoM/discussions/1399

<div type='discussions-op-text'>

<sup>Originally posted by **IsakNaslundBh** June  9, 2022</sup>
The `IElementM` currently have two required methods, `MaterialComposition` and `SolidVolume`, as outlined in https://github.com/BHoM/documentation/wiki/IElement-required-extension-methods .

For a lot of cases though, the volume needs to be computed as part of evaluating the Material composition, to ensure the correct ratios are computed. This can therefore be wasteful, and make the computation harder than it would be so simple return a material takeoff, with `volume : Material` pairs.

Suggesting that the requirements on the `IElementM` are changed to be that you have to do one or both of the options below:

1. Current requirement. Implement SolidVolume and MaterialComposition
2. Implement a MaterialTakeoff method that returns a class as described above (similar to MaterialComposition, but with volumes instead of rations of total volume)

It would be trivial for the MatterEngine to go from one too the other by simple addition/division, and this way the developer could implement whichever of the two is making most sense for the particular object.

You could ofc also implement both on a particular object, as long as you ensure that the results fully overlap for all cases.

Happy to hear thoughts from the group.

@JosefTaylor @michaelhoehn @FraserGreenroyd @enarhi @al-fisher @peterjamesnugent </div>

IsakNaslundBh:

Closes #1422 

Adds class for MaterialTakeoff. Class for storing a set of materials and the corresponding volumes of said materials.

### DIFF
--- a/Physical_oM/Materials/MaterialTakeoff.cs
+++ b/Physical_oM/Materials/MaterialTakeoff.cs
@@ -1,0 +1,63 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace BH.oM.Physical.Materials
+{
+    [Description("Defines the make up of an object through a list of Materials and their corresponding volumes." +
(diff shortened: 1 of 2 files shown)


pawelbaran (changes requested):
Please note this is to large extent a duplicate of https://github.com/BHoM/Revit_Toolkit/blob/main/Revit_oM/Misc/RevitMaterialTakeOff.cs - shouldn't the existing class be migrated or replaced?

IsakNaslundBh:
> Please note this is to large extent a duplicate of https://github.com/BHoM/Revit_Toolkit/blob/main/Revit_oM/Misc/RevitMaterialTakeOff.cs - shouldn't the existing class be migrated or replaced?

Happy to migrate and version the existing revit class into this new one, as long as it does not have massive repercussions meaning this PR turns into to big of an alignment exercise. If the latter is the case, we can raise an issue and get it handled in a later PR push (this milestone). Are you fine with that @pawelbaran ?

@michaelhoehn I think you have used the pre-existing class the most, maybe you can chip in to say if it can safely be replaced by this new class or not (minding the slight difference in class structure, effectively doing the same thing).

pawelbaran:
Sounds good to me, as long as we do not leave both items living separate lives in long term 👍

IsakNaslundBh:
> Sounds good to me, as long as we do not leave both items living separate lives in long term 👍

Realised that https://github.com/BHoM/Revit_Toolkit/blob/main/Revit_Engine/Query/MaterialTakeOff.cs is 100% interfering with the current strategy, so will have to make a tweak anyway. Will put do-not-merge tags on this until we have a Revit solution up.

Might require a bit more discussion before that is actioned though, but thanks for flagging this @pawelbaran as it actually turned out to be a problem that needs to be fixed now.

IsakNaslundBh:
@BHoMBot check compliance

alelom (approved):
Code change looks good. Adding do-not-merge until I finish reviewing/testing connected PRs.

IsakNaslundBh:
@BHoMBot check core
@BHoMBot check serialisation

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1441, opened 2022-10-20
New DisplayText attribute

### LINKED ISSUE #1440: Provide a centralised solution for setting a display text on enum values
#### Description:
Currently enum values are shown as they are defined in the code. For example, Central Europe will show as `CentralEurope` since we cannot have spaces in an enum name. 

To solve that, I have been using the `Description` attribute to store a UI friendly text representation along with my own methods to convert enums to and from text. As the number of cases for this solution grows, it would be better to switch to a centralised solution.

It is however important to not that, within the BHoM, the `Description` attribute has already been used on enums to provide actual descriptions and cannot therefore be used to contain the display text. 

image

I suggest to use a new attribute: `DisplayTextAttribute` that will serve that single purpose. The `ParseEnum` method in the engine will need to be updated to support that new attribute. We will also need to update teh ToText method so it now looks for the `DisplayText` attribute on enum values.

adecler:

Closes #1440 

See issue for description.

### Test files
See Engine PR: https://github.com/BHoM/BHoM_Engine/pull/2930

### DIFF
--- a/BHoM/Attributes/DisplayTextAttribute.cs
+++ b/BHoM/Attributes/DisplayTextAttribute.cs
@@ -1,0 +1,57 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd (approved):
See [this review](https://github.com/BHoM/BHoM_Engine/pull/2930#pullrequestreview-1149405802)

adecler:
@BHoMBot check core
@BHoMBot check serialisation

adecler:
@BHoMBot check ready-to-merge

travispotterBH:
@adecler Can you please add this to the wiki on attributes describing its use?

adecler:
@travispotterBH , Done.

PR #1472, opened 2023-02-06
BHoM: Add attribute for documentation URL attribute

FraserGreenroyd:

Fixes https://github.com/BHoM/BHoM_Engine/issues/471


FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check compliance

BHoM/Attributes/Enums/DocumentationType.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Text;
+
+namespace BH.oM.Base.Attributes.Enums
+{
+    public enum DocumentationType
+    {
+        Undefined,
+        APILink,
+        SampleScript,
+        WikiLink,
```
alelom, line 34: A few opinions here:
- Since the switch to MkDocs we need to move away from the name "Wiki" I think -- rather say `Docs` or `Documentation` link
- I would not add the name `link` in the enum values:
    -  it should be clear from the content you provide
    -  it should be consistent across entries (e.g. you should also say `SampleScriptLink`, or rename the enum `DocumentationTypeLink` and get rid with `Link` in the entries)
- Do we need Undefined only for UI reasons? This enum is reflected in the UI, but I think 99% of the cases will be used from the code. `Undefined` feels redundant/error-prone -- you either provide a Documentation attribute with a valid enum value or not

FraserGreenroyd: I've removed the `Link` word from the options, and renamed `Wiki` to `Documentation`.

`Undefined` is always on my enums as a default option - it will help the bot filter out problems down the line so I'd rather keep it in line with expected enum values elsewhere.

alelom (approved):
Thanks for addressing my comments. Looks good!

FraserGreenroyd:
@BHoMBot check installer
@BHoMBot check core
@BHoMBot check serialisation
@BHoMBot check ready-to-merge

PR #1478, opened 2023-02-09
Updating enum name

### LINKED ISSUE #1211: Structure_oM: Misleading parameter name and description in LoadingPanelProperty
#### What is missing/incorrect?

`LoadingPanelProperty` features a parameter titled `LoadApplication`, whose description says "Load application enum [...]"

This is misleading as the enum type name is actually `LoadPanelSupportConditions`, making it quite difficult to find the correct enum.

image

kprusicka:

Closes #1211
Enum name was updated.

### DIFF
--- a/Structure_oM/SurfaceProperties/Enums/LoadApplication.cs
+++ b/Structure_oM/SurfaceProperties/Enums/LoadApplication.cs
@@ -23,7 +23,7 @@
 namespace BH.oM.Structure.SurfaceProperties
 {
     /***************************************************/
-    public enum LoadPanelSupportConditions
+    public enum LoadApplication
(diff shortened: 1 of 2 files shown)


FraserGreenroyd (changes requested):
This PR would require versioning support, and updating a host of dependant toolkits. A good addition/attempt in the hack session, but requires more thought/work to resolve the issue, which @IsakNaslundBh is not convinced is needed following an offline chat with him yesterday. As such, this PR will not go for deployment, and I am closing the PR as it won't be merged in its current state. @IsakNaslundBh I will keep the branch in case you do decide you want to do something with it.

PR #1485, opened 2023-02-28
Structure_oM: adding a new DistanceIsParametric `bool` property to `BarPointLoad`

### LINKED ISSUE #1427: Structure_oM: `BarPointLoad.DistanceFromA` Description attribute needs to clarify whether distance is parametric
#### What is missing/incorrect?

I had this question asked from an external collaborator -- we need to specify whether the distance is intended as parametric or actual in the attribute description. After confirming with @peterjamesnugent , adapters assume that this is parametric (as it is also for other loads, [e.g. `BarDifferentialTemperatureLoad.TemperatureProfile`](https://github.com/BHoM/BHoM/blob/95d6557547f2bedb1c5262cb06772be67781244f/Structure_oM/Loads/BarDifferentialTemperatureLoad.cs#L43-L44)).

https://github.com/BHoM/BHoM/blob/95d6557547f2bedb1c5262cb06772be67781244f/Structure_oM/Loads/BarPointLoad.cs#L39-L40

alelom:

Closes #1427

Adds a new `DistanceIsParametric` boolean to `BarPointLoad` to clarify whether `DistanceFromA` property is parametric or absolute. Adjusts the description to match.

### Test files

No workflows support this yet. Done in support of https://github.com/BHoM/Karamba3D_Toolkit/pull/4#discussion_r1116882312.

### DIFF
--- a/Structure_oM/Loads/BarPointLoad.cs
+++ b/Structure_oM/Loads/BarPointLoad.cs
@@ -36,8 +36,12 @@ namespace BH.oM.Structure.Loads
         /***************************************************/
 
         [Length]
-        [Description("Distance along the Bar between the StartNode and the load position.")]
+        [Description("Distance along the Bar between the StartNode and the load position." +
+            "If `DistanceIsParametric` is true, this is a number between 0 and 1 defining the parametric distance along the Bar.")]
         public virtual double DistanceFromA { get; set; } = 0;
+
+        [Description("Whether the DistanceFromA is to be intended as absolute or parametric. Defaults to false.")]
+        public virtual bool DistanceIsParametric { get; set; } = false;
 
         [Force]
         [Description("Magnitude and direction of the Force. The load requires the Force and/or the Moment Vector to be non-zero to have any effect.")]


IsakNaslundBh:
Putting a do-not-merge tag on this, as I do not think we should merge this until all adapters currently supporting this load type has been adjusted to either:

1. Support this change, and be able to handle it
2. Raise errors telling the user it is not yet supported.

Without this, this can make for a situation where you create a load in BHoM that is expected as one thing, but comes with different meaning in the software it is pushed to.

Also, need to align methods like the Visualize method in the Structure_Engine to accurately display the loads.

FraserGreenroyd:
@alelom @IsakNaslundBh is there an update on this PR and the work needed to close it out? Keen to ensure we have adequate resource to complete as necessary.

IsakNaslundBh:
@FraserGreenroyd There is quite a bit of work still outstanding, namely, some updates to engine methods, such as the visualisation,  and above that, updates to all toolkits to make sure they can handle this new property.

Think we also should update the BarVaryingDistributedLoad with the same boolean once we decide to go ahead with this PR.

FraserGreenroyd:
@IsakNaslundBh @alelom any further update to this in the last month? Conscious of your impending (planned and short term!) departure @IsakNaslundBh and wouldn't want anything hanging unnecessarily? 😄

alelom:
After discussion with @IsakNaslundBh , it appears that the work for this PR is quite extensive as it requires a wider refactoring in several places. If we add this feature here, we also need to align BarVaryingDistributedLoads, the Create methods in Structure, the Visualisation methods, and all toolkits will need to be able to correctly push and check for this type of load.

Because the initial reason behind this was the development around Karamba, which has now reached a stable point, there is no priority to push this further. With @IsakNaslundBh we've both agreed that this is a feature we'd need to introduce, but a wider workshop/refactoring effort needs to be planned in order for it to be carried out.

PR #1511, opened 2023-05-02
Base_oM: Input classification attributes refactored

### LINKED ISSUE #1507: BHoM: Enable passing arguments to InputClassificationAttribute constructor
#### Description:

We would like to introduce `InputClassificationAttribute` with properties to enable e.g. file path inputs with enforced file extension. To achieve this, we need `InputAttribute` constructors that would pass `params object[]` to the constructor of `InputClassificationAttribute`.

### LINKED ISSUE #1508: Add FileExtension property to FilePathAttribute
#### Description:

As in the title, there is a request for file path inputs with enforced file extension, which can be achieved by simply adding `FileExtension` property to `FilePathAttribute`.

pawelbaran:

Closes #1507 - different approach taken, but the issue will disappear
Closes #1508

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%2FBase%5FoM%2F%231511%2DFileExtensionPropertyOnFilePathAttribute&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) - open the script and check if all inputs and outputs have appropriate descriptions. Test object and method to be deleted before the PRs get merged.

### Changelog

- `InputClassificationAttribute` renamed to `ClassificationAttribute` (it is used for classifying both inputs and outputs), `Name` property added
- `InputAttribute` refactored to accept only `Quantity` types in constructor - all other classifications (`FilePath`, `FolderPath`) to exist as independent attributes to be added in separate declarations, i.e.:
    ```C#
    [Description("length", "Length.", typeof(Length))]
    public static void MethodAcceptingLength(double length)
    {
    }

    [FilePath("filePath")]
    [Description("filePath", "File path.")]
    public static void MethodAcceptingFilePath(string filePath)
    {
    }
    ```
- `QuantityAttribute` migrated from Quantity_oM to BHoM (required to compile after limiting the constructor of `InputAttribute` to accept only `QuantityAttribute`)
- `FileExtensions` property added to `FilePathAttribute` to allow for limiting the acceptable file extensions
- constructor of `ClassificationAttribute` taking `int` as name added to target multi outputs - the solution with casting to string is a bit sketchy, but seems to be secure (method parameters cannot have names starting from digits)
- an assumption is made that there can be only one classification attribute per input/property - quantities, paths etc. will never be combined

### Additional comments

This PR is the outcome of the discussion between @adecler, @IsakNaslundBh and myself, which dragged me away from the solution adding `params object[]` constructor parameter to be passed into `Activator.CreateInstance` of `InputClassificationAttribute`. I believe I got to the point where I researched possible solutions and chose the one that looks most appropriate - sounds like a good moment for code review to me.

The PR is not ready to merge - refactoring in dependent repos will be required, but I would like to get the feedback on the chosen solution first. If positive, I will proceed with alignment of other repos, add versioning and remove test object and method.


FraserGreenroyd (changes requested):
From discussion, limit this PR to only upgrading `FilePathAttribute` and remove the inheritance to `InputClassificationAttribute` - update any repos needed for that and handle the `InputClassificationAttribute` changes to a future PR to be picked up in the future.

BHoM/Attributes/ClassificationAttribute.cs
```diff
@@ -28,8 +28,39 @@
...
+        public ClassificationAttribute(string name)
+        {
+            Name = name;
+        }
+
+        /***************************************************/
+
+        public ClassificationAttribute(int index) : this(index.ToString())
+        {
+
```
FraserGreenroyd, line 60: If we could add a code comment explaining the `int index` is for the `MultiOutput` usecase that would beneficial for the next time we come back to this.

FraserGreenroyd (commented):
Following call with @pawelbaran we have worked out and agreed on the following:

 - `InputClassificationAttribute` to be renamed to `ClassificationAttribute` as proposed - this makes more sense given it is also used on `Output` attributes!
 - `Name` property being added to `ClassificationAttribute` to support mapping attributes which inherit from `ClassificationAttribute` to their method parameters
 - `FilePath` and `FolderPath` attributes to be fleshed out as proposed currently, providing more functionality and flexibility to them that doesn't exist currently

This will aim to resolve this PR and the current needs that this PR was raised under. Going forward though we will seek to have a broader discussion on attributes in the next milestone with a view to tidy up and align all attributes and clearly define attributes and inheritance. As such, no changes to the inheritance will be made in this PR to limit its scope.

PR #1545, opened 2023-08-14
Structure_oM: Add classes for Foundation objects

### LINKED ISSUE #1543: Structure_oM: Add classes for foundations
#### Description:

Add foundation classes for `PadFoundation` and `PileFoundation` with an overarching `IFoundation` interface. 

@michaelhoehn, @mchaf, @pawelbaran, @enarhi tagging you here as you are all involved in the Physical_oM discussions. @Martian42, @Maria1129 and @Chrisshort92 from the ground engineering aspect. 

This should be coordinated with the Physical_oM where the following issues are open:
https://github.com/BHoM/BHoM/issues/566
https://github.com/BHoM/BHoM/issues/1155
Foundation Classes

The rationale:
- The `IFoundation` makes the implementation in to more complex objects easier (the only commonality is the `OrientationAngle`;
- `PadFoundation` is primarily defined by a `PlanarSurface` to give flexibility on the shape, we can write Engine methods to have simpler creators;
- `ConstantThickness` used to limit the `Property`, but by implementing `IAreaElement` we could change this `Property` (this would give us access to all the area loads etc. and would rely on the users not to assign `Waffle` properties to a `PileCap`;
- It is possible to have variable depth foundations, for example the use of shear keys;
- `PileCap` is just a `PadFoundation` as they are effectively the same, solid objects (usually reinforced concrete);
- `PileLayout` is defined by an `ExplicitLayout` and then Engine methods can orientate this using the `OrientationAngle` to a specific point;
- I do not think a `Pile` needs a specific object as it would just contain the `Lengh` and `PileSection` - but that may be useful for the Physical objects?

peterjamesnugent:

Closes #1543 

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM/Structure_oM/%231543-FoundationClasses/Foundation%20classes%20and%20associated%20methods.gh?csf=1&web=1&e=Q17tVv

### Changelog

- Added classes for `PadFoundation`, `PileFoundation` and `PileGroup` as representation of foundation objects
- Added `IFoundation` as the encompassing interface for `PadFoundation` and `PileFoundation`

### Additional comments

- Changed `PadFoundation` to have `List<Edge>` instead of `ISectionProperty` to avoid conflict of `Material` between that and the `IProperty` - also to allow the `IElement2D` requirements to be satisfied.
- `IProperty` used so that `IAreaElement` can be implemented and we get the loads for free (we can limited this to `ConstantThickness` in the `Create` methods;
- `PileFoundation` inherits the `IElementM` to allow use of the `VolumetricTakeOff` methods;


peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

Structure_oM/Elements/PadFoundation.cs
```diff
@@ -0,0 +1,29 @@
...
+    [Description("2D element for structural analysis. The PadFoundation is a planar surface object defined by an IProfile and a ISurfaceProperty.")]
+    public class PadFoundation : BHoMObject, IFoundation, IAreaElement, IElement2D
+    {
+        [Description("The profile used to define the outline of the pad.")]
+        public virtual List<Edge> TopSurface { get; set; }
+
+        [Description("Defines the thickness property and material of the PadFoundation.")]
+        public virtual ISurfaceProperty Property { get; set; } = null;
+
+        [Description("Local x, y, and z axes of the PadFoundation as a vector Basis. Defaults to null which is interpretated to defaults when pushed to software and world axes in BHoM.")]
```
pawelbaran, line 25: You are the native speaker here, please feel free to ignore 😉 
```suggestion
        [Description("Local x, y, and z axes of the PadFoundation as a vector Basis. Defaults to null which is interpreted to defaults when pushed to software and world axes in BHoM.")]
```

Structure_oM/Elements/PileGroup.cs
```diff
@@ -0,0 +1,30 @@
...
+using BH.oM.Structure.SurfaceProperties;
+using BH.oM.Dimensional;
+using BH.oM.Quantities.Attributes;
+using BH.oM.Spatial.Layouts;
+using BH.oM.Spatial.ShapeProfiles;
+using BH.oM.Structure.SectionProperties;
+
+namespace BH.oM.Structure.Elements
+{
+    public class PileGroup : BHoMObject, IFoundation
```
pawelbaran, line 18: Description is missing

Structure_oM/Elements/PileGroup.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        [Description("The length of the piles within the PileGroup.")]
+        public virtual double PileLength { get; set; }
```
pawelbaran, line 21: Worth adding `[Length]` attribute?

Structure_oM/Elements/PileGroup.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        [Description("Section property of the PileGroup, containing all sectional constants and material as well as profile geometry and dimensions, where applicable.")]
+        public virtual ISectionProperty PileSection { get; set; }
+
+        [Description("Pile Layout defining the position of the piles about the World Origin.")]
```
pawelbaran, line 26: This description could be improved I believe - how does the layout specified here correspond to the coordinate system of the cap?

Structure_oM/Elements/PileFoundation.cs
```diff
@@ -0,0 +1,26 @@
...
+using BH.oM.Geometry;
+using BH.oM.Structure.SurfaceProperties;
+using BH.oM.Dimensional;
+using BH.oM.Quantities.Attributes;
+using BH.oM.Structure.SectionProperties;
+using BH.oM.Spatial.Layouts;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("PileFoundation is a composite object consisting of a pile cap and a group of piles.")]
```
pawelbaran, line 17: As in the other comment, I believe it is important to explain the relationship between the coordinates of the cap and piles

Structure_oM/Elements/PileFoundation.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class PileFoundation : BHoMObject, IFoundation, IElementM
+    {
+        [Description("The pile cap defined as a PadFoundation.")]
+        public virtual PadFoundation PileCap { get; set; }
+
+        [Description("Pile Layout")]
```
pawelbaran, line 23: This could be a bit more precise 😉

Structure_oM/Elements/PadFoundation.cs
```diff
@@ -0,0 +1,29 @@
...
+using BH.oM.Quantities.Attributes;
+using BH.oM.Spatial.ShapeProfiles;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("2D element for structural analysis. The PadFoundation is a planar surface object defined by an IProfile and a ISurfaceProperty.")]
+    public class PadFoundation : BHoMObject, IFoundation, IAreaElement, IElement2D
+    {
+        [Description("The profile used to define the outline of the pad.")]
+        public virtual List<Edge> TopSurface { get; set; }
```
pawelbaran, line 20: One more thought: will you ever make any use of release properties of `Edge`? If not, maybe the geometrical representation could be represented by a `PolyCurve` or similar?

pawelbaran (commented):
Probably not too surprisingly, I will take the role of a poor advisor here - leaving more questions than answers 🙈 On top of the ones left in the code comments, I have two more thoughts worth vocalising I believe:
- Will we ever need to consider piles as individual entities, e.g. to do the structural analysis of each member etc.? Would be worth to estimate the future need for `Pile` object.
- The current design is a bit different compared to what has been discussed during the recent class design brainstorming session (`List<Edge>` instead of `IProfile`), which results in decoupling of the piles from their caps - now `PileFoundation` consists of a `PadFoundation` and a collection of `PileGroups`, each floating independently in the global coordinate system. I am not saying it is incorrect, but sounds a bit inconsistent to me - easy to go out of sync, feels less intuitive to do the calcs, also low ability to share geometrical characteristics within groups of foundations.

pawelbaran:
Morning @peterjamesnugent, I have 3 more thoughts following yesterday's review:
- with the current setup of `PileFoundation` class, wouldn't it be more practical to have `List<Pile>` to represent piles instead of `List<PileGroup>`? Cap edges are independent of the local coordinate system, piles in group are represented by individual points in the global coordinate system too, so what's the value in grouping them? Following our offline chats - it is all pros and cons, so not suggesting anything, rathe food for thought 👍 
- if we stick to `PileGroups`, why wouldn't we use `ILayout2D` as a property defining their layout? It costs nothing and I can imagine it being potentially helpful when optioneering different pile layouts?
- it would be great if `PileFoundation` implemented one of `IElementXD` interfaces - otherwise it does not get the support of all the great stuff that we have implemented for dimensional elements, some of which being quite necessary, such as `Transform` (required e.g. to pull from Revit linked models correctly), so you will need to implement such methods manually to make sure the object behaves as expected

Cheers!

Structure_oM/Elements/PadFoundation.cs
```diff
@@ -0,0 +1,29 @@
...
+using BH.oM.Base;
+using BH.oM.Geometry;
+using BH.oM.Structure.SurfaceProperties;
+using BH.oM.Dimensional;
+using BH.oM.Quantities.Attributes;
+using BH.oM.Spatial.ShapeProfiles;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("2D element for structural analysis. The PadFoundation is a planar surface object defined by an IProfile and a ISurfaceProperty.")]
```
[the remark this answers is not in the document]
peterjamesnugent: Have updated to `    [Description("2D element for structural analysis representing a pad foundation. The PadFoundation is a planar surface object defined by edges and an ISurfaceProperty.")]`

Structure_oM/Elements/PadFoundation.cs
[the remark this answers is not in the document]
peterjamesnugent: Agree if it were a `Panel` but for pad foundations and pile caps you don't really use the centre of the pad - you typically define the top level plus a thickness and go from there.

Structure_oM/Elements/PileGroup.cs
peterjamesnugent: Agree - updated to:

"Pile Layout defining the position of the piles about the World Origin. when used with the PileFoundation object, the piles will be located to the centre  of the pile using the Basis provided by the PileCap."

peterjamesnugent:
@BHoMBot check required

Structure_oM/Elements/Pile.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using BH.oM.Structure.SectionProperties;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("A pile object defined by it's geometry and section.")]
+    public class Pile : BHoMObject, IFoundation, IElement1D, IElementM
+    {
+        [Description("Defines the top Node. Note that Nodes can contain Supports.")]
+        public virtual Node TopNode { get; set; }
```
pawelbaran, line 22: I guess the decision to use nodes instead of `Line` property has been thought through, but will still let myself ask an ignorant question 😉 Do releases or supports at endpoints of a pile bear any meaning? I would be a bit afraid of confusing the engineers by leaving space to set properties that are then ignored.

Structure_oM/Elements/Pile.cs
```diff
@@ -0,0 +1,41 @@
...
+using BH.oM.Base.Attributes;
+using BH.oM.Base.Attributes.Enums;
+using BH.oM.Dimensional;
+using BH.oM.Geometry;
+
+using BH.oM.Structure.SectionProperties;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("A pile object defined by it's geometry and section.")]
```
pawelbaran, line 18:
```suggestion
    [Description("A pile object defined by it's location and section.")]
```

Structure_oM/Elements/PileFoundation.cs
```diff
@@ -0,0 +1,26 @@
...
+using BH.oM.Structure.SurfaceProperties;
+using BH.oM.Dimensional;
+using BH.oM.Quantities.Attributes;
+using BH.oM.Structure.SectionProperties;
+using BH.oM.Spatial.Layouts;
+
+namespace BH.oM.Structure.Elements
+{
+    [Description("A composite object representing a pile foundation. This object contains a pile cap and list of piles which can be used for structural analysis.")]
+    public class PileFoundation : BHoMObject, IFoundation, IElementM
```
pawelbaran, line 18: I am thinking of potential implementation of `IElement` interface here - as mentioned in the offline chat at some point, it really helps to automate the location handling. However, it feels like none of the interfaces that we currently have actually applies to this object.

If we zoom out a bit, it is a composite object consisting of `IElement2D` and a collection of `IElement1D`. Maybe we should think of implementing an `ICompositeElement` interface that would let us access and interact with its child objects of types that `IElement`?

Ofc not for this PR, just dumping my thoughts here.

FraserGreenroyd:
@BHoMBot check documentation-compliance

Structure_oM/Elements/Pile.cs
peterjamesnugent: Valid questions, when modelling piles or pile groups for analysis you typically apply a support at the endpoints e.g. a pin - so I don't think the `Constraint6DOF` owned by the `Node` will be disregarded.

Structure_oM/Elements/PadFoundation.cs
peterjamesnugent: Updated to `PolyCurve`.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:

peterjamesnugent:
@BHoMBot check required

Chrisshort92 (approved):
Functionality and terminology tested with test script attached. Pr is approved

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:

peterjamesnugent:
@BHoMBot check versioning

peterjamesnugent:
@BHoMBot check versioning

PR #1547, opened 2023-08-16
Diffing_oM: null coalescence of optional Diff constructor parameters added

### LINKED ISSUE #1546: Diffing_oM: introduce null coalescence of optional parameters in the Diff constructor
#### Description:

Currently `Diff` constructor has 2 optional parameters defaulted to null. This causes null reference issues downstream, which could easily be resolved by adding null coalescence, as discussed with @alelom.
https://github.com/BHoM/BHoM/blob/fbde3edb1ba298e4efa26b11fa13a43fac7c27c1/Diffing_oM/Diff.cs#L68-L76

pawelbaran:

Closes #1546

### Test files

Standard test procedure for diffing.

### DIFF
--- a/Diffing_oM/Diff.cs
+++ b/Diffing_oM/Diff.cs
@@ -71,8 +71,8 @@ namespace BH.oM.Diffing
             RemovedObjects = removedObjects;
             ModifiedObjects = modifiedObjects;
             DiffingConfig = diffingConfig;
-            ModifiedObjectsDifferences = modifiedObjectsDifferences;
-            UnchangedObjects = unchangedObjects;
+            ModifiedObjectsDifferences = modifiedObjectsDifferences ?? new List<ObjectDifferences>();
+            UnchangedObjects = unchangedObjects ?? new List<object>();
         }
 
         /***************************************************/


pawelbaran:
@BHoMBot check required

alelom (approved):
Approving based on code review and discussion with @pawelbaran .

The reason why we can add the null coalescing on the last 2 props (`ModifiedObjectsDifferences` and `UnchangedObjects`) is that they are _optional_ (but still _defining_, not derivable) in the definition of a Diff object. Instead, the first 4 properties are non-optional, and if any of them is null, all the others lose sense; for this reason, I asked Pawel not to add null coalescing on them. The responsibility on the correct `Diff` instantiation and the null checking on the first 4 properties must be owned by the downstream code that instantiates the `Diff`. The downstream code must be responsible for the `Diff` consistency, instantiate it "at the very last moment", and feed its ctor non-null, consistent values.

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: versioning, ready-to-merge

PR #1632, opened 2024-08-08
Geometry_oM: Changes for refactor of polyline offset

IsakNaslundBh:

Adding OffsetOptions class to help with new and improved full refactor of the offset for Polylines


pawelbaran (commented):
After reviewing the engine, I have a feeling that only the 1st property is indeed relevant. First of all, I am not a big fan of keeping mutually exclusive settings under one class, i.e. some properties are relevant _only_ for polycurves, while the others _only_ for polylines.

Even more importantly, all properties but first are only used for polyline offset - and why would polyline offset yield different results than a curve that contains an arc of massive radius? Also I do not think it brings any practical value to be able to keep invalid segments etc.

IsakNaslundBh:
@pawelbaran I have updated the descriptions, correcting the description for HandleGeneralSelfIntersection and added comment about the only valid reason I can see to not have it on (change in topology).

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

pawelbaran (approved):
Happy to approve following a thorough review in collaboration with @IsakNaslundBh.

PR #1641, opened 2024-10-07
LifeCycleAssessment_oM: Refactor LCA_oM to run on dynamic objects and add Transport support

IsakNaslundBh:

https://github.com/BHoM/LifeCycleAssessment_Toolkit/issues/306

Overall refactor of all the base classes in the toolkit to make use of the new DynamicObejcts, to allow flexibility on which modules to store and report results for.

Add support for explicit transport factors for A4 and C2 phases, through the CombinedLifeCycleAssessmentFactors class and the three new ITransportFactors classes.

All to be tested together with https://github.com/BHoM/LifeCycleAssessment_Toolkit/pull/357

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM/LifeCycleAssessment_oM/%231641-AddCalculatedMaterialLifeCycleEnvironmentalImpactFactors?csf=1&web=1&e=3l4XIN

### Changelog

- Refactor metrics and results to dynamic objects
- Add new single factor metrics
- Add support for transport factors
- Add new CombinedFactors class to handle a combination of EPD and transport
- Clean up enum names
  - EnvironmentalMetrics -> MetricType
  - LifeCycleAssessmentPhases -> Module
- Add Combination and sub-modules to the module enum

### Additional comments

The number of files changes and lines changed might look scarier than it is, due to there being quite a bit of copy paste across all the metrics. Can generally divide by 17, as that is the number of metrics we so far have added to the oM.


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check compliance

adecler (commented):
Looking at this PR from a code architecture perspective, Here's a few comments/questions for discussion:

 - Typo in the name: `ILifeCycleAssemsmentIndicator`
 - Do we need those interfaces: `IMaterialResult` & `IElementResult<out T>`? We already have the abstract classes that are the only one implementing it.
 - ModuleDescriptions are only used in one place so why not have them directly on the Enum?
 - You have a folder for interfaces but most of the interfaces are found in other folders
 -  Not sure why `SingleTransportModeImpact` and `VehicleEmissions` are not just one class
 - Things that need clean-up/upgrade?
   - ProjectLifeCycleAssessment  	--> only the elements it contains matter 
   - LifeCycleAssessmentScope    	--> Not used anywhere
   - HealthProductDeclaration    	--> Only appears when converted from custom object
   - BuildingBenchmarkingData    	--> Not sure about this one. It's used in `BenchmarkVariance` method but I don't really see the point of that method. Also appear in conversion from custom object.
   - AdditionalEPDData		--> Not sure about that one either. It's very rigid set of properties for something that's labelled as `Additional`. Not really used anywhere so wonder if there is a point keeping it here. If used in another toolkit, seems better to move it there.
   - EPDDensity			--> Appreciate this is kept for backward compatibility but wonder if we shouldn't rip the bandaid. That old workflow was bad so keeping old schemas is just creating noise and confusion
   - Scope			--> It's another fragment so feels like it's a candiadate for deletion but actually used so just need to understand what it does better.

PR #1644, opened 2024-10-22
Verification_oM added

### LINKED ISSUE #1643: Add Verification oM and Engine
#### Description:

As discussed with @al-fisher and @IsakNaslundBh, we are going to add projects containing code that would enable verifying objects and object sets against certain requirements.

pawelbaran:

Closes #1643

### Test files

See https://github.com/BHoM/BHoM_Engine/pull/3431


pawelbaran:
@al-fisher @alelom @adecler please be informed this PR is ready for review. I did my best to address all comments made by @IsakNaslundBh, I believe we landed at really good agreement but I would still encourage you to have a close look to make sure you feel comfortable with what is going to be merged.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check code-compliance

pawelbaran:
btw @adecler @alelom did the compliance check go funky or am I missing a point?

michal-pekacki (approved):
Approving after internal discussion with @pawelbaran and testing via https://github.com/BHoM/BHoM_Engine/pull/3431

pawelbaran:
Thanks @michal-pekacki. @adecler @alelom @al-fisher I would like to merge this PR by the end of tomorrow if no other comments are made - please have a look at this PR by then in case you would like to contribute 👍 Thanks!

pawelbaran:
@BHoMBot check core
@BHoMBot check versioning
@BHoMBot check serialisation

al-fisher (approved):
Thanks @pawelbaran 
I am happy to add a high level approving review as looking through your implementation it is certainly broadly aligned with the strategic plan as discussed and agreed! That is to implement this capability as this new foundational Verification oM.
Pleased to see this progressing. Well done!
Thanks

Verification_oM/Conditions/HasId.cs
```diff
@@ -0,0 +1,40 @@
...
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace BH.oM.Verification.Conditions
+{
+    [Description("Condition that requires an object to own an AdapterId fragment with a given Id to be considered a pass.")]
+    public class HasId : ICondition
```
alelom, line 29: Just a comment here
- Because you use the name of what this concept "has", I think this type is most suited as an _interface_ called `IHasId` or `IWithId` or similar
- It can be also a _class_, but then the general rule should be to call it with a name of what it is, rather than what it has or should have/implement: classes represent concepts, [not contracts](https://cseducators.stackexchange.com/q/544/15440). So `ConditionWithId` or similar

pawelbaran: Good points @alelom, thank you. I actually had a bit of a hard time with this particular class as I inherited it from your old day's prototype, which actually is a condition class 😉
https://github.com/CIHDigitalConfigurator/CIH_Toolkit/blob/master/CIH_oM/Conditions/HasId.cs

Now trying to understand your current perspective, I am not sure why would 'HasId' be considered as a contract - yes, it defines a behaviour 'has', but it is not the behaviour of the object represented by that particular object, it defines the expected behaviour of an object checked against that condition, so it is a slightly different case than usual.
Also although the requested property (the collection of AdapterIds) is not itself strictly related to a `ICondition`, `ICondition` interface explains that this class represents a condition, thus it is there.

I have a feeling it may boil down to semantics and naming rather than class structure, but then I see a range of other cases to look at:
- `IsInDomain`
- `IsInSet`
- `IsOfType`
- `IsNotNull`

Just thinking about it, would adding `Condition` suffix to the name solve the problem?

alelom: I see. I think I was reasoning thinking that `HasId` was supposed to be a parent type that can be implemented by downstream conditions. Instead, it is properly a Condition that can be applied on its own. See `details` below as per why I got confused (TL;DR: due to my rough architecture in the old CIH toolkit).

In short, I think this PR is fine to keep as is. We could make it clearer by making all Condition classes `sealed`, and/or by making them inherit a parent `public abstract class BaseCondition : ICondition`, rather than directly implementing the `ICondition` interface, see below. I don't think either one is a reason to keep this PR open, so feel free to go ahead and merge.

<details>

I got mistaken because, in my old architecture, I had only created an `ICondition` interface, but not a base `BaseCondition` parent class. The thing is, if we have `HasId`, `IsInDomain`, etc., and these only implement the `ICondition` interface, then it's not too clear whether they are conditions themselves, or if they simply have to respect the contract of the `ICondition` interface. 

For clarity, we could have a base abstract class `BaseCondition` that every condition needs to inherit, much like `BHoMObject`, so there's no confusion. Then, every downstream condition class can inherit from the `BaseCondition` and be called without the `...Condition` suffix. If dowstream classes only implement the `ICondition` interface, you have no certainty they are conditions themselves. 

Further, because the properties in each Condition class are supposed to be well designed to define a self-enclosed condition that can be applied on its own, we could make them `sealed`, so we clarify that they can't be parent types to other conditions.

Therefore, we could think of creating this parent class:  
`public abstract class BaseCondition : ICondition`  
and then have every concrete Condition class be a `sealed` class that only implements `BaseCondition`, e.g.  
`public sealed class HasId : BaseCondition`

This is mostly a suggestion for clarity and I would not hold this PR for this.

</details>

pawelbaran: Thanks @alelom, all makes sense and I actually agree with you. My [original design](https://github.com/BHoM/BHoM/pull/1644/commits/9868be03472ef48a0d4a05a4d7769e5b435a9b48) contained `BaseCondition`, but I got convinced by @IsakNaslundBh that the core oM classes should be as simple as possible, i.e. no inheritance, even from a base class (which I do not fully agree with in this particular case). So I am definitely not saying no, to this particular comment, but would be good to have @IsakNaslundBh back before we continue.

I raised https://github.com/BHoM/BHoM/issues/1646 to capture this discussion, to be revisited later in this milestone to apply necessary changes before the code gets released. Hopefully that lets us mark this thread as resolved and continue with merging this PR 👍

pawelbaran:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 32722694611

PR #1677, opened 2025-04-15
Add support for dynamic objects

### LINKED ISSUE #1665: Add support for dynamic objects
#### Description:

As the Life Cycle Assessment Toolkit is being used in an increasing number of contexts, it became clear that the specific LCA modules that need to be stored on an object vary significantly depending on the region, discipline and reporting entity.

Currently, each module is stored as a separate property. This causes a few problems:
 - We regularly have to add properties to cover new modules. This makes the schemas unstable and constantly increasing in size. The users end up facing large UI components with many inputs/outputs irrelevant for their context.
 - Modules are often grouped and represented by a single environmental result (e.g A1-A3). This means the schemas also must cover combinations of modules, resulting in multiple create methods that must guarantee combined modules are auto calculated. 

With the current state of the framework, storing those module results in a list or a dictionary would reduce the quality of the user experience by:
 - The module results would not appear as clear inputs and outputs of components, making it more cumbersome for the user the input or extract the relevant information for specific modules.
 - Loss of the tooltips clearly indicating the units of each metric.

In summary, we would like each module result to be dynamically available depending on context while retaining all the I convenience of actual properties

adecler:

Closes #1665

This adds support for Dynamic objects, including their handling of quantity attributes. 
This offers 3 options to create dynamic objects with UI support in the BHoM:
 - Option 1: The most flexible. The new objects will have to implement the `IDynamicPropertyProvider` interface and provide extension methods for `GetProperty`, `SetProperty`, and `GetProperties`.
 - Option 2: This allows to have objects with dynamic properties of different quantity types without having to implement extension methods explicitly.  The class definition needs to have a dictionary with enum keys and values of type `IQuantity`. Plus the `DynamicPropertyAttribute` on that dictionary. The class can have multiple dictionaries of that sort.
 - Option 3. Like option 2 but with all dynamic properties in a dictionary are of the same quantity. The dictionary then has values of type `double` and the quantity attribute is set directly on the dictionary.

image

image

Check the [Removing test classes](https://github.com/BHoM/BHoM/commit/fcc3d548e81017b091add22745146e88f2444097) commit to see example of the 3 options for the case of the LCA toolkit. 

### Test files
[DynamicObjects_Test.zip](https://github.com/user-attachments/files/19796444/DynamicObjects_Test.zip)
This script needs to be opened in Rhino 8. Ideally, you would test this using your own project where dynamic objects are needed.

Needs to be tested alongside those 2 PRs:
 - https://github.com/BHoM/BHoM_Engine/pull/3492
 - https://github.com/BHoM/BHoM_UI/pull/517

### Additional comments

It is worth noting that this is currently addressing the main issue of providing full support for dynamic objects in the UI but there's a few more aspects that would be good to consider as we move forward:
 - Hash 
 - Diffing
 - Reflection methods like `PropertyDictionary` and `PropertyNames`
 - Serialisation

Do we want the dynamic properties to be treated as regular properties only in the UI or do they also need to be elevated in other, less exposed aspects like serialisation and diffing ?


adecler:
@BHoMBot check compliance

adecler:
@BHoMBot check required

BHoM/Interface/IDynamicObject.cs
```diff
@@ -0,0 +1,38 @@
...
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System.ComponentModel;
+
+namespace BH.oM.Base
+{
+    [Description("Identifies objects that have dyanmic properties.")]
```
IsakNaslundBh, line 27:
```suggestion
    [Description("Identifies objects that have dynamic properties.")]
```

Quantities_oM/IQuantity.cs
```diff
@@ -0,0 +1,37 @@
...
+using BH.oM.Base;
+using BH.oM.Quantities.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text;
+
+namespace BH.oM.Quantities
+{
+    [Description("Represents a numerical value with an SI unit attached to it.")]
```
IsakNaslundBh, line 32: Think we should say that it relates to a type of quantity rather than unit. Not 100% on this, but at least think so.

```suggestion
    [Description("Represents a numerical value with a quantity type attached to it.")]
```

Quantities_oM/Quantity.cs
```diff
@@ -0,0 +1,51 @@
...
+using BH.oM.Quantities.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Drawing;
+using System.Text;
+
+namespace BH.oM.Quantities
+{
+    [Description("Represents a numerical value with an SI unit attached to it.")]
```
IsakNaslundBh, line 33:
```suggestion
    [Description("Represents a numerical value with a quantity type attached to it.")]
```

IsakNaslundBh (commented):
Some description comments

PR #1679, opened 2025-04-22
Fix compilation via dotnet build

### LINKED ISSUE #1678: The code doesn't compile when using dotnet build
#### Description:
As per title.  No project should have `GenerateSerializationAssemblies` set to `On`

#### Steps to reproduce:
Run the command as below: 

Image

#### Expected behaviour:
`GenerateSerializationAssemblies` is not needed for the BHoM as we are not using the [XmlSerialiser](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/ide/reference/build-page-project-designer-csharp?view=vs-2015&redirectedfrom=MSDN). So we should turn it off.

#### Test file(s):
No need for test file, just run `dotnet build .\BHoM.sln` in a command window.

adecler:

Closes #1678

Two projects files had `GenerateSerializationAssemblies` turned on so both were fixed.

### Test files
No need for test file, just run `dotnet build .\BHoM.sln` in a command window to make sure the code now compiles successfully.

### DIFF
--- a/BHoM/BHoM.csproj
+++ b/BHoM/BHoM.csproj
@@ -9,7 +9,7 @@
     <FileVersion>8.2.0.0</FileVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <GenerateSerializationAssemblies>On</GenerateSerializationAssemblies>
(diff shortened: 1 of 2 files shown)


michaelhoehn (approved):
Works perfectly!

adecler:
@BHoMBot check ready-to-merge

PR #1686, opened 2025-06-04
Refactoring of the versioning toolkit

adecler:
### NOTE: Supported PR
https://github.com/BHoM/Versioning_Toolkit/pull/302

   
See PR above for details

### DIFF
--- a/Versioning_oM/Attributes/DowngraderAttribute.cs
+++ b/Versioning_oM/Attributes/DowngraderAttribute.cs
@@ -1,0 +1,66 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2025, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Base.Debugging;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


Versioning_oM/Attributes/VersioningTargetAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace BH.oM.Versioning
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    [Description("Attribute defining the specfic type or method a versioning method is targetting. This attribute can be used multiple times if a method is targetting more than one element.")]
```
IsakNaslundBh, line 36: Don't think this is used for methods? I have at least never seen it used for methods, and looking at the `UpgradeMethod` is has no references to the `ToNewObject` that stores the upgrade methods. Guess we could change that if we like, but given the `PreviousVersionAttributes` already works well, I see no real need.

Also, added a minor comment about that the type name need to be fully namespace qualified.

```suggestion
    [Description("Attribute defining the specific type a versioning method is targeting. The target type needs to be specified including the full namespace. This attribute can be used multiple times if a method is targeting more than one element. ")]
```

Versioning_oM/Attributes/UpgraderAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace BH.oM.Versioning
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [Description("Attribute identifying a class as an upgrader for versioning.")]
```
IsakNaslundBh, line 36: Might be worth adding something here specifying that the version numbers should be the version you are upgrading to and not from. That if you are working in 8.1, this should be 8.1, not 8.0.

```suggestion
    [Description("Attribute identifying a class as an upgrader for versioning. The version specified should be the the version that the object is being upgraded to which generally correspond to the current version of the BHoM.")]
```

Not sure if my suggested change is clear enough.

Versioning_oM/Attributes/DowngraderAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace BH.oM.Versioning
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [Description("Attribute identifying a class as a downgrader for versioning.")]
```
IsakNaslundBh, line 36: Same as with the Upgrader.
```suggestion
    [Description("Attribute identifying a class as a downgrader for versioning.  The version specified should be the the version that the object is being downgraded from which generally correspond to the current version of the BHoM.")]
```

IsakNaslundBh (changes requested):
Some comments on descriptions, trying to make them even clearer. Happy for my suggestions to be changed

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #1717, opened 2026-04-29
Structure_oM: Added design result object

### LINKED ISSUE #1716: Structure_oM: Add DesignResult Object
#### Description:

Add a design result object that stores the governing action, the corresponding resistance, and the required factor of safety for generic structural analysis.

MariaGlimaker:

Closes #1716

### Changelog

- Added `IDesignResult` class as a basis for all code specific design results;


MariaGlimaker:
Add unit as a QuantityType https://github.com/BHoM/BHoM/blob/7a681296a235ede723b3b372d0a6be799ed53c7c/LifeCycleAssessment_oM/Enums/QuantityType.cs#L25

Structure_oM/Results/Design Results/DesignResult.cs
```diff
@@ -0,0 +1,130 @@
...
+        public virtual int ModeNumber { get; }
+
+        [Description("Time step for time history results.")]
+        public virtual double TimeStep { get; }
+
+        [Description("The structural resistance or capacity of the element, representing the maximum load or stress the element can withstand.")]
+        public virtual double Resistance { get; }
+
+        [Description("The design action or demand acting on the element, representing the applied load or stress effect.")]
+        public virtual double Action { get; }
```
peterjamesnugent:
```suggestion
        [Description("The structural resistance or capacity of the element, representing the maximum load or stress the element can withstand.")]
        public virtual double Utilisation{ get; }
```

Structure_oM/Results/Design Results/DesignResult.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        [Description("The minimum required factor of safety as specified by the applicable design standard.")]
+        public virtual double RequiredFactorOfSafety { get; }
```
peterjamesnugent:
```suggestion
```

Structure_oM/Results/Design Results/DesignResult.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        [Description("Human-readable description of the design check, explaining what is being verified.")]
+        public virtual string Description { get; }
+
+        [Description("Specifies the type of design check, defining how Action and Resistance are evaluated.")]
+        public virtual DesignCheckType CheckType { get; }
```
peterjamesnugent:
```suggestion
```

Structure_oM/Results/Design Results/DesignResult.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        [Description("Messages generated during the design check to communicate assumptions made, inapplicable checks, out-ofrange values, or missing data.")]
+        public virtual List<string> Feedback { get; }
```
peterjamesnugent:
```suggestion

```
Not on abstract class

peterjamesnugent:
@BHoMBot check core
@BHoMBot check compliance

Structure_oM/Design/IDesignResult.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.ComponentModel;
+using BH.oM.Analytical.Results;
+using BH.oM.Base;
+using BH.oM.Structure.Results;
+
+namespace BH.oM.Structure.Design
+{
+    [Description("Base class for structural design check results that can be implemented for specific design codes and actions.")]
+    public abstract class IDesignResult : IStructuralResult, IImmutable, IResultItem
```
IsakNaslundBh:
```suggestion
    public abstract class DesignResult : IStructuralResult, IImmutable, IResultItem
```

This is not an interface, hence don't think it should have the leading I there.

Structure_oM/Design/IDesignResult.cs
```diff
@@ -0,0 +1,109 @@
...
+        public virtual IComparable ResultCase { get; }
+
+        [Description("Positive index, starting at one. Only set for cases with modal outputs such as dynamic cases.")]
+        public virtual int ModeNumber { get; }
+
+        [Description("Time step for time history results.")]
+        public virtual double TimeStep { get; }
+
+        [Description("The utilisation of the design check.")]
+        public virtual double Utilisation { get; }
```
IsakNaslundBh:
```suggestion
        public abstract double Utilisation { get; }
```

Would consider making this abstract to force the subclasses to explicitly implement it.

Think it would be good to force all results to have this defined on them and clearly visible. WOuld also force all subclasses to have a proper description for the particular utilisation.

IsakNaslundBh (approved):
After our discussions, I am happy with this to be tested out.

Like the modularity of it. Just have to check if the duplication of data (for all identifiers) becomes an issue and if we want to switch to larger bulk classes instead, but in principle I like this as a concept, so lets go for it and see if we need to adjust later.

Only small comment on the class name (should not have the leading I given it is a class and not an interface.)

Also comment on the Utilisation property as I think it would be best if that is forced to be overriden, so the most critical property is not hidden in this baseclass.

Also scope to later add specific base classes for things like Bars etc (containing position as well). This could be done as a separate PR though.

IsakNaslundBh (dismissed):
ABove review was supposed to be request for change.

peterjamesnugent:
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh (approved):
Happy with this now given - all comments in previous review still pending.

Not sure if we need to worry about the failing check or not, but given this is a greenfield new class I do not really see how and why it should impact versioning.

To be sure I made a code search for `DesignResult` and there are no other classes with that name that I could find.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot copyright-compliance

peterjamesnugent:

sakanni:
Good to merge- red versioning check is a known CI issue, not a problem with this PR. Full sweep of serialised objects misattributed to repo due to namespace matching. Non-required check so it doesn't block merge

peterjamesnugent:
@BHoMBot check ready-to-merge

